PR #2, opened 2018-10-16
Header component

jleach:
Initial work on header component.

This header is done in legacy non-mobile first design so we can get one component shipped. I'll refactor the header to be mobile-first in a subsequent PR.


rajpersram:
I've no problem with the code, but on a related note moving forward we should discuss two items:
1) Font sources. I recommend using Google Fonts API, referenced remotely so we don't have to install fonts into our project - the latter unfortunately requires at least 3 types of font types per font.
2) Let's decide and use a SASS structure for our CSS so we're not writing CSS by hand. See this note in our Taiga Wiki for a suggested setup: https://taiga.pathfinder.gov.bc.ca/project/lukegonis-developergovbcca/wiki/css-implementation

patricksimonian:
how is this component going to be injested by the markdown file that is to present this content? Also I'm with @rajpersram on using Google Fonts API if that is allowed. From what I've checked though they don't have **Myriad Web Pro** as that is a Adobe licensed font?

jleach:
@patricksimonian The html / css for the commonest are just samples. I'll actually copy the specific CSS and HTML (for the header) to the markdown file that will be (we hope) picked up by Gatsby. I might just delete the samples when done.

ksingbeil:
Looking good! 
A couple things:
Positioning of logo/title - should be more indented. Logo should link back to gov.bc.ca
Logo - switch logo to sunrise with BC beside (not below)
Font - Site title and copy should be in Myriad Pro...I know there is a separate discussion about needing
 a licensed font for (supposedly open) government work...
Changing the logo will make it possible to make the header tighter (it's a bit high now).
Visually, the header should resemble this (more/less, no nav) - https://portal.nrs.gov.bc.ca/web/client/home

jleach:
@ksingbeil The only logo in that format is pretty low quality with the blue background embedded. It would be idea to get the logo in `svg` format. You can see in the NR link that the logo is blurry. Know where we can get a better copy?

Let's talk about the use of `Myriad Pro` we have an open and free one that looks almost identical. Its also hosted by google so much faster load times and it can be optimized. I think Myriad Pro might be legacy from printed material.

rajpersram:
@ksingbeil btw, the font mentioned is a Google Font called PT Sans.

ksingbeil:
Thanks @jleach and @rajpersram . We're chatting with the branding folks this week and will raise these points - and find a better logo!

jleach:
@orinevares Are you able to add your initial work on the harder markdown to this branch / RP? When we have the html / css dialled I'll port it to that file and we can go over it with @patricksimonian to make sure its something he can import.

orinevares:
@jleach no problem, ill upload the file now.

jleach:
@orinevares @ksingbeil I've updated the header to look more like the NR header.

@rajpersram I've had to hard code some values in size. For example, the heard is fixed at 65px (no media queries in place just yet). And I've also put a 65px left and right border on each side. Your thoughts?

Check out the two images I put in the screenshots folder.

jleach:
@orinevares FYI to make the header to edge to edge the `body` needs to have the margins corrected to 0 with `<body style="margin: 0;">`. This brought up the question of included some thoughts on a reset in the design system.

rajpersram:
@jleach Further to our talk about a universal way to reset html tags via css, the original normalize.css (from MIT) was incorporated into Bootstrap and Google Material Design also uses this (with some sass changes). With that in mind, I suggest in general using the BootstrapCDN's reset css, called reboot.css, found here: https://stackpath.bootstrapcdn.com/bootstrap/4.1.3/css/bootstrap-reboot.min.css.

components/header/README.md
```diff
@@ -0,0 +1,123 @@
+---
+description: Headers help users understand what the content of the page is about and provides a quick, organized way to reach the main sections of a website.
+---
+# Header
+
+Headers help users understand what the content of the page is about and provide a quick, organized way to reach the main sections of a website. They appear at the top of a page, above the main body text.
+
+### Required
+### Last Updated: 2 days ago
```
patricksimonian: Last updated should probably be a definitive date instead of something relative otherwise no matter what the user would perceive this to have been updated recently.

orinevares:
@jleach 

Things are looking good, below are a few changes to the header.

1. The gold line should be a bit thicker, to mimic the NR site
2. The site title text should be a bit bigger, according to the developer guide that is 42 pts
3. When the browser window width decreases the site title should disappear instead of the words stacking on top of one another. Also just like the NR site.

jleach:
The gold like is 2px. I think it looks smaller on our page because of the white below it. I also:

Increased the size to ~35pt to match the NR site;
Tweaked the media queries to hide the text at a certain size;
Cleaned up the blurry logo a bit.

jleach:
@orinevares @ksingbeil Are one of you able to approve the PR? There should be a button in the top right to start a review of the changes. Make comments in-line if needed, then approve or request changes. LMK if you need a walk through.

ksingbeil:
Useful document on the use of the BC Logo - 
http://www.corporate.gov.bc.ca/print-ads/Govt_of_BC_Logos/Resources/Third-Party_Use_of_BCMarks.pdf

orinevares:
@jleach just approved it, let me know if there are any issues with it

PR #11, opened 2018-10-23
Build out footer component

jleach:
Initial work on footer component.

This footer is done in legacy non-mobile first design so we can get one component shipped. I'll refactor the footer to be mobile-first in a subsequent PR.


orinevares:
Looking good.

One fix:
When hovering over the title it should underline but not turn blue.

components/footer/style.css
```diff
@@ -0,0 +1,34 @@
+footer {
+  background-color: #036;
+  border-top: 2px solid #fcba19;
+  color: #fff;
+  flex-shrink: 0;
```
patricksimonian: flex shrink not needed here. It only works for flex children

PR #22, opened 2018-11-15
Implement navigation menu in CSS and HTML

jleach:
Implement the navigation (component) mechanics that accompany the header. Also, make the header sticky for a more modern look and feel.


jleach:
@patricksimonian This is my initial work on the nav button and feedback thus far is appreciated.

jleach:
@orinevares @ksingbeil I've added the "button" for the navigation bar. A few things for consideration are: 

First, I reduced the left and right margin from 65px to 30px; 65 was a bit too big. It looks like the NRS site we like is even less. I can reduce it further if needed. 

Second, I used `FontAwsome` for the icon (the NRS site uses it as well). It'll be in the same category as Fonts and boiler plate CSS in that it is hosed on a CDN (at the moment).

Third, I made the header sticky. While its not in the NRS site after some discussion we though it was a more modern approach. It takes us from 2007 to 2014 in terms of web design :)

components/navbar/style.css
```diff
@@ -0,0 +1,80 @@
...
+    right: 10px;
+    z-index: 100;
+    width: auto;
+    margin: 0;
+    /* padding: 5px 8px 4px;
+    /* border-color: #fcba19;
+    background: #036; */
+  }
+
+  header a [class^="fas fa-"] {
```
jleach: @patricksimonian Not sure if this was needed but if I understand it correctly it'll style any FA icons in the `header`. Overkill?

ShellyXueHan:
remember to add the visual(screenshot) in readme when finalized

components/navbar/style.css
patricksimonian: i think it  may be better left off to the implementers since they may choose to use a different icon library like bootstrap icons or even FontAwesome's svg icon library

orinevares:
@jleach thanks for your work on this!

My only feedback right now would be to match the margins of the BC logo and site title to the original basic header. They should both be identical except that one has the nav bar built in.

components/navbar/sample.html
```diff
@@ -0,0 +1,53 @@
...
+    <div class="other">
+      <a class="nav-btn">
+        <i class="fas fa-bars"></i>
+      </a>
+    <!-- 
+      This place is for anything that needs to be right aligned
+      beside the logo.
+    -->
+    </div>
+    </div>
```
patricksimonian: extra div

patricksimonian: is there going to be a seperate component for the `<nav>..</nav>`

components/navbar/style.css
```diff
@@ -0,0 +1,81 @@
+header {
+    background-color: #036;
+    border-bottom: 2px solid #fcba19;
+    padding: 0 30px 0 30px;
+    color: #fff;
+    display: flex;
+    height: 65px;
+    top: 0px;
```
patricksimonian: as a practice, if the value is 0px the 'px' part is usually dropped `0px` -> `0`

components/navbar/style.css
```diff
@@ -0,0 +1,81 @@
...
+    position: fixed;
+    width: 100%;
```
patricksimonian: I think we probably won't need to set the width explicitly if the `.other` container has the `flex-grow: 1` property

components/navbar/style.css
```diff
@@ -0,0 +1,81 @@
...
+    -webkit-box-shadow: 0 6px 8px -4px #b3b1b3;
+    -moz-box-shadow: 0 6px 8px -4px #b3b1b3;
+    box-shadow: 0 6px 8px -4px #b3b1b3;
+  }
+  
+  header h1 {
+    font-family: 'PT Sans';
+    font-weight: normal;  /* 400 */
+    margin: 5px 5px 0 18px;
+    visibility: hidden;
```
patricksimonian: is the h1 title dropped?

components/navbar/style.css
```diff
@@ -0,0 +1,81 @@
...
+    flex-grow: 1;
+    /* border-style: dotted;
+    border-width: 1px;
+    border-color: lightgrey; */
+  }
+
+  header .nav-btn {
+    top: 15px;
+    right: 10px;
+    z-index: 100;
```
patricksimonian: I may be wrong here, but i think `top, right, z-index` have no effect here since this is not a `positioned` element. 

top/right only works on positioned elements `position: relative, position: absolute, position: fixed` etc.

components/navbar/style.css
```diff
@@ -0,0 +1,81 @@
...
+    }
+  }
+  
+  @media screen and (min-width: 900px) {
+    header h1 {
+      font-size: 2.0em;
+      visibility: visible;
+    }
+  }
+
```
patricksimonian: new line

components/navbar/sample.html
```diff
@@ -0,0 +1,53 @@
...
+  </p><p style="margin-left: 35px; margin-right: 35px;">
+    Sed consectetur egestas lectus sed hendrerit. Donec molestie orci quis erat scelerisque aliquet. Sed auctor maximus feugiat. Proin mollis, risus at dapibus dapibus, dolor metus volutpat sapien, sed condimentum magna turpis non purus. Pellentesque auctor felis eget rhoncus pellentesque. Sed neque enim, imperdiet vitae egestas placerat, commodo sed mauris. Etiam egestas lacus ac ex commodo laoreet. Suspendisse sed augue vitae ex elementum hendrerit. Pellentesque ullamcorper facilisis augue, ut lacinia ipsum consectetur sit amet. Aenean eget diam ac neque scelerisque placerat.
+  </p>
+</body>
+
+</html>
```
patricksimonian: new line

components/navbar/sample.html
```diff
@@ -0,0 +1,53 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <link rel="stylesheet" href="https://use.fontawesome.com/releases/v5.5.0/css/all.css" integrity="sha384-B4dIYHKNBt8Bc12p+WXckhzcICo0wtJAoU8YZTY5qE0Id1GSseTk6S+L3BlXeVIU" crossorigin="anonymous">
+    <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/twitter-bootstrap/4.1.3/css/bootstrap-reboot.min.css">
+    <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=PT Sans">
+    <link rel="stylesheet" href="style.css">
```
patricksimonian: what do you think on including the standard set of meta tags as apart of the head?
```html
<head>
   <meta charset="utf-8"/>
   <meta http-equiv="x-ua-compatible" content="ie=edge"/>
   <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no"/>
</head>
```

components/navbar/sample.html
jleach: Fixed. Also, I'm going to add the nav to this component next.

components/navbar/style.css
jleach: Fixed.

components/navbar/style.css
jleach: Fixed.

components/navbar/style.css
jleach: Fixed.

components/navbar/sample.html
jleach: Fixed.

components/navbar/style.css
jleach: Fixed.

components/navbar/sample.html
jleach: Added.

components/navbar/style.css
jleach: Its hidden for small screens but visible otherwise.

components/navbar/style.css
jleach: Do you think I should be doing this with the SVG icon set?

jleach:
I've added the navigation and an animated GIF to illustrate the look and feel.

orinevares:
@jleach thank you, looks good to me! 
@ksingbeil or @bl-smith any edits?

ksingbeil:
@jleach Looking good! Thanks :)
A couple things:
- Could the hamburger menu not have a box around it (like Gov.bc.ca) and be white?
- Is there extra space below the nav text? Seems like there is extra padding. https://engage.gov.bc.ca/govtogetherbc/ is an example that has tighter padding below the nav text.

bl-smith:
Hey Jason - thanks for all this work! I love the animated gif - that's super cool, and I think will really help people on the design system get a sense of the intended behaviour of the component.  

One question: I see you're pulling from Font Awesome for the hamburger icon. I think that's great, but am wondering if pulling from Font Awesome raises any of the same security concerns that a google font might? Would it be safer just for us to download the SVG from the font awesome site, then provide it as an asset that can be downloaded from the Design System page?  Not sure what makes most sense, but just wanted to raise that. Thanks!

jleach:
I made a few tweaks. Check it out.

components/navbar/sample.html
```diff
@@ -0,0 +1,97 @@
...
+      if (window.getComputedStyle(elem).display === 'block') {
+        hide(elem);
+        return;
+      }
+
+      show(elem);
+    };
+
+    // Listen for click events
+    document.addEventListener('click', (event) => {
```
patricksimonian: was there a reason to bind this globally?
i'd change this to specifically bind to an element instead of the entire document.
```
const menuButton = document.getElementById('menu');
menuButton.addEventListener('click', ... 
```

components/navbar/sample.html
```diff
@@ -0,0 +1,97 @@
...
+      // Filter events
+      if (event.target.id !== "menu") return;
```
patricksimonian: use of double quotes mixed with single quotes

jleach:
It does. Whatever we solution for the typeface we’ll apply to FA and the Bootstrap ‘reboot’ component. I’ll fix all of them at the same time.

J

components/navbar/sample.html
```diff
@@ -0,0 +1,97 @@
...
+  </p><p style="margin-left: 25px; margin-right: 25px;">
+    Sed consectetur egestas lectus sed hendrerit. Donec molestie orci quis erat scelerisque aliquet. Sed auctor maximus feugiat. Proin mollis, risus at dapibus dapibus, dolor metus volutpat sapien, sed condimentum magna turpis non purus. Pellentesque auctor felis eget rhoncus pellentesque. Sed neque enim, imperdiet vitae egestas placerat, commodo sed mauris. Etiam egestas lacus ac ex commodo laoreet. Suspendisse sed augue vitae ex elementum hendrerit. Pellentesque ullamcorper facilisis augue, ut lacinia ipsum consectetur sit amet. Aenean eget diam ac neque scelerisque placerat.
+  </p>
+
+  <script>
+    const useCapture = false;
```
patricksimonian: over all looks really good. Only suggestion is to wrap this entire thing into a function so that these variables are not set globally

patricksimonian: also maybe use ES5 javascript since this is a browser specific example? unless we provide comments explaining that this code should be transpiled

components/navbar/sample.html
jleach: Fixed.

components/navbar/sample.html
jleach: Fixed.

PR #33, opened 2018-11-21
Change official gov URL to use the canonical name www2

danpollock:


jleach:
@danpollock Thanks for the PR. We're going to look into the history of http://www.gov.bc.ca vs http://www2.gov.bc.ca/.

danpollock:
@lukegonis, @jleach : any word on why this was rejected?

PR #34, opened 2018-11-22
Convert badge to png

patricksimonian:
# About

SVG files are not fetchable in the devhub and so the badges appear as broken images. 

I converted the SVG files into PNG's as a work around.
.https://taiga.pathfinder.gov.bc.ca/project/lukegonis-developergovbcca/issue/684


jleach:
What makes them non-fetchable?

patricksimonian:
@jleach from what I've read so far, svg files are unable to request external resources.
comment from [github markup issue 556](https://github.com/github/markup/issues/556)
```
@kivikakk That's absolutely correct (based on my own research into this). Referencing an SVG using an <img> tag rasterizes the image and prevents it from further loading anything. It's basically inert.

Here's the most relevant literature I know about what SVG features are available when using the <img> tag: https://www.smashingmagazine.com/2014/11/styling-and-animating-svgs-with-css/#embedding-svgs

This is the key clause:

Styles and animations applied to an SVG using an external CSS resource will not be preserved once the SVG is embedded.

Basically, the SVG can't make any external references.
```

PR #37, opened 2018-11-30
Weekly Update to Design System

orinevares:
For @jleach to review
- Adding HTML, CSS, and Markdown for text input component

Other changes:
- Updates to primary button, checkbox, radio button, and Nav bar code
- Adding documentation from Taiga into Github


jleach:
@patricksimonian Did you want to check out this PR? You'll spot HTML / CSS issues that I won't.

components/about/component_workflow.md
```diff
@@ -0,0 +1,36 @@
+---
+description: Components are constantly being added to the design system. The backlog shows what’s on the list, upcoming, and currently in development.
+title: Component Backlog
+status: draft
+---
```
patricksimonian: are these documents meant to be apart of devhub?

components/text_input/README.md
```diff
@@ -0,0 +1,81 @@
+---
+description: Text inputs allow users to enter any combination of letters, numbers, or symbols of their choosing in a structured format.
+title: Text Input
+status: Draft
+---
```
patricksimonian: add author attributes (should be your github username)

components/text_input/sample.html
```diff
@@ -0,0 +1,19 @@
...
+  <head>
+    <meta charset="utf-8">
+    <link rel="stylesheet" href="./style.css">
+    <link href="https://fonts.googleapis.com/css?family=PT+Sans" rel="stylesheet">
+    <title>Text Input</title>
+  </head>
+  <body>
+
+    <form>
+      First Name:<br>
```
patricksimonian: replace with `<label>First Name:</label>`

components/about/component_workflow.md
orinevares: I believe these will eventually go under the about section of the design system. @bl-smith can't remember what you decided last week about making these public on DevHub or just keeping them on GitHub.

components/text_input/sample.html
patricksimonian: this may make the labels show to the left of the input field, if that occurs one way around it is to surround
the label, input into a div and flex it.

components/about/component_workflow.md
bl-smith: Thanks for looking at this, @patricksimonian  - the devhubignore file does have the line "docs/*.md" - doesn't that meant that if we put the Change Request Process file in that folder it will be ignored? The other two files should get pulled into Dev Hub, so no worries there.

patricksimonian: yes it will :)  but this file appears to be in /components/... right?

patricksimonian: I may be having the case of the mondays

orinevares: Ok makes sense.

I can add the ignore:true to the front matter of the change request process.

patricksimonian: ohh i misread my bad. Anyhow..bryan if you could add `author: bl-smith` to the front matter of the docs you've written. We have added support for avatars to the devhub. It'll help tart the whole thing up.

components/text_input/sample.html
orinevares: Done

components/text_input/README.md
orinevares: ok will do now

components/about/component_workflow.md
bl-smith: @orinevares  and @ksingbeil - maybe we could talk further about how we want our avatars to appear and where. I notice that the GovUK uses their generic Github account (govuk-design-system) to publish documentation on Github. This makes sense to me, because if someone has a question about the design system processes or administration, they should contact the Design System Administrators, not necessarily me. Should we do something similar?

patricksimonian: @orinevares @bl-smith @ksingbeil good points all, if you would prefer the avatar to be just bcgov. The `author: orinevares` attribute (for example) can just be switched to `author: bcgov`

jleach: The images is open to the public via GitHub. If people don't want it to be public than for consistency it should be removed from GitHub and replaced with a non-personal avatar. Otherwise it should be left in.

ksingbeil: @patricksimonian BCGov feels too generic. I think a Design System avatar would be good for documentation...but I'm not sure the logistics on how this connects with Git - does this mean there needs to be a generic Design System Git account? I think this treads into what @jleach is talking about.

patricksimonian: Hmm, this is actually leading to a support issue on how to 'display an avatar' in devhub. The technical aspects right now is that we are assuming the 'author' is a github user, that way we can easily grab the avatar.

So I'll have to answer the questions in two ways.

First yes, techinically *at this time* a design system github account would be needed, the 'author' attributes for all markdown files would have to be changed

Second possible answer, this is actually begging the question for the ability to add a avatar image in the configuration. No promises on that as it requires a bit of research into some web standards on what can be [Unfurled](https://medium.com/slack-developer-blog/everything-you-ever-wanted-to-know-about-unfurling-but-were-afraid-to-ask-or-how-to-make-your-e64b4bb9254) from content.

orinevares: Hi All,

Seems like this thread has taken a new direction and raised a further conversation. For the sake of closing this PR I will mark this as resolved but with the understanding to investigate the avatar feature further.

PR #38, opened 2018-11-30
add author attribute to markdown files

patricksimonian:


ksingbeil:
@patricksimonian Sure! Sounds good.

PR #41, opened 2018-12-07
Weekly Update to Design System

orinevares:
- Updating markdown files to match results of user testing and contain additional content
- Updating code to include Noto Sans and accessibility testing results


components/checkbox/README.md
```diff
@@ -8,66 +8,67 @@ author: orinevares
 ![Status](https://img.shields.io/badge/Component-Draft-orange.svg)
 
 # Checkbox
-Checkboxes are a form of control inputs that allow users to select one or more options from a list.
+Checkboxes are a form of control inputs that allow users to select one or more items from a list.
 
 ## Required
 
-## Last Updated: 2 days ago
+## Last Updated:
```
patricksimonian: are we not providing updates?

components/footer/style.css
```diff
@@ -2,6 +2,7 @@ footer {
   background-color: #036;
   border-top: 2px solid #fcba19;
   color: #fff;
+  font-family: 'Noto Sans','Calibri', 'Arial', 'Sans Serif';
```
patricksimonian: liking the fallback font families a lot 👍

components/checkbox/README.md
orinevares: Yes, just removing the placeholder text for now

components/navbar/README.md
```diff
@@ -120,6 +121,52 @@ This component builds on the `header` component; See `sample.html` for a working
...
+  </p><p style="margin-left: 25px; margin-right: 25px;">
+    Sed consectetur egestas lectus sed hendrerit. Donec molestie orci quis erat scelerisque aliquet. Sed auctor maximus feugiat. Proin mollis, risus at dapibus dapibus, dolor metus volutpat sapien, sed condimentum magna turpis non purus. Pellentesque auctor felis eget rhoncus pellentesque. Sed neque enim, imperdiet vitae egestas placerat, commodo sed mauris. Etiam egestas lacus ac ex commodo laoreet. Suspendisse sed augue vitae ex elementum hendrerit. Pellentesque ullamcorper facilisis augue, ut lacinia ipsum consectetur sit amet. Aenean eget diam ac neque scelerisque placerat.
+  </p>
+
+  <script>
+    function setupSampleMenuControl() {
```
patricksimonian: i'd suggest leaving a javascript comment here explaining that this is a sample implementation and that your implementation will vary based on your front end framework/techstack

components/text_input/README.md
```diff
@@ -103,12 +99,12 @@ This library is updated on a regular basis; by subscribing to the design library
 ### CSS
 ```
 form {
-  font-family: 'PT Sans','Calibri', 'Arial', 'Sans Serif';
+  font-family: 'Noto Sans','Calibri', 'Arial', 'Sans Serif';;
```
patricksimonian: i don't think font family is necessary here. Ideally we'd have that set in a top level element such as the body

patricksimonian: that or provide a comment explaining that we are using font family here for the example display purposes, but ideally you'd be setting this up in a higher page element.

orinevares: It seemed to not affect the text within the input field but I could be wrong

components/navbar/README.md
orinevares: Ok

components/checkbox/README.md
```diff
@@ -8,66 +8,67 @@ author: orinevares
...
+
+### Screenreaders
+* Proper ARIA and ALT labels are included
+* Language tag has been set to english
+* List items and checkbox status are labelled properly
+
+### Colour Contrast
+* Contrast ratios are at least 7:1 for normal text and 4.5:1 for large text
+
+### Tab Navigation
```
bl-smith: Can we change this to keyboard navigation instead of Tab navigation?

components/checkbox/README.md
```diff
@@ -8,66 +8,67 @@ author: orinevares
...
+## Prototype with This Component
 Download and use the B.C. government Design System library to help build prototypes and wireframes for your website or application.
 
 The library is constantly updated and currently available in the following tools:
 
-•	Sketch: [https://sketch.cloud/s/Q0bkG] 
-
-This library is updated on a regular basis; by subscribing to the design library, you’ll be informed of updates to components and new additions.
+*	[Sketch](https://sketch.cloud/s/Q0bkG)
+* [UxPin and others](https://sketch.cloud/s/Q0bkG)
```
bl-smith: I like this like to UXPin and others, but wonder if it could be confusing to just link to the same Sketch library. Can we maybe create a page explaining how to integrate sketch files into UXpin and other tools, and which tools are currently compatible?

components/checkbox/README.md
orinevares: Sure, that makes more sense.

components/footer/README.md
```diff
@@ -12,68 +12,55 @@ Footers help people find what they need after scrolling to the bottom of a web p
...
+This checkbox has been built according to [WCAG 2.0 AA](https://www.w3.org/TR/WCAG20/) standards and all government services should strive to meet this level.  This component successfully includes the following accessibility features:
 
-1. Perceivable
-*	Alternative text is provided for non-text content (WCAG 1.1.1)
-*	Colour is not used as the only visual means of conveying information (WCAG 1.4.1)
-*	Visual presentation of text and images has a contrast ration of at least 4.5:1 (WCAG 1.4.3)
-*	Text can be resized without assistive technology up to 200% without loss of content or functionality (WCAG 1.4.4)
-*	Images of text are only used for pure decoration (WCAG 1.4.9)
+### Screenreaders
+* Proper ARIA and ALT labels are included
```
bl-smith: Have we included ARIA labels yet? I agree that we should, but not sure if this is incorporated in our components yet.

components/text_input/README.md
```diff
@@ -9,69 +9,65 @@ author: orinevares
...
+* Input text that is no longer than a single line
 
+## Don't Use This when
+* A user needs to input multiple lines of text. Use a textarea component instead
+* Users can only choose from a selected list of answers
 
+## Design Protocol
+* The length of the text input should provide a hint to the user on how much text to write.
+* Text inputs must have visible labels aligned above the text input they refer to. They should be short, direct, and written in sentence case.
+* Allow copy and paste functionality
```
bl-smith: "Allow copy and paste functionality."  - add a period at the end so that it is consistent with punctuation on the other lists

components/checkbox/README.md
```diff
@@ -8,66 +8,67 @@ author: orinevares
...
-* If a user can only select one option from a list, use a radio button instead.
-
 ## Rationale
 Our checkboxes are larger than most browser defaults to allow for greater visibility and larger target areas for touchscreen devices. Research has shown that users resort to only clicking the checkbox and not the associated text, therefore a larger target area is necessary. Simply enlarging browser defaults proved to be problematic on many web browsers and the custom design should be used.
-Based on research done by Gov.UK: https://designnotes.blog.gov.uk/2016/11/30/weve-updated-the-radios-and-checkboxes-on-gov-uk/
+Based on [research](https://designnotes.blog.gov.uk/2016/11/30/weve-updated-the-radios-and-checkboxes-on-gov-uk/) done by Gov.UK.
 
 ## Behaviour
 1. Users should be able to tap or click on either the checkbox or the label to select the option.
+2. Do not preselect options, this signals a biased result.
```
bl-smith: Comma splice - change to: "Do not preselect options. This signals a biased result."

components/checkbox/README.md
patricksimonian: gotchya

components/checkbox/README.md
orinevares: Ya I agree we need an explanation page. The UxPin link is really to just convey an initial understanding that the sketch library works with other tools. I'll take out the link for now until we have a page or process outlining how this is done.

PR #81, opened 2019-01-30
Accessibility Updates

orinevares:


components/primary_button/README.md
```diff
@@ -35,22 +35,15 @@ The rounded edges on the button help differentiate it from other layout items on
...
-* Proper ALT labels are included
-* Language tag has been set to english
+In addition to [fundamental accessibility]() requirements, this component meets or exceeds the [WCAG 2.0 AA](https://www.w3.org/TR/WCAG20/) standards by incorporating the following criteria.
 
 ### Colour Contrast
-* Contrast ratios are at least 7:1 for normal text and 4.5:1 for large text
-
-### Keyboard Navigation
-* Supports tab navigation for use without a mouse
+* Contrast ratios are at least 7:1 for normal text and 4.5:1 for large text when in both hover and static states
```
dlevineBC: It's not clear what is meant by normal text vs large text. Isn't this just stating the contrast ratio for either normal state vs hover state?

components/text_input/README.md
```diff
@@ -37,19 +37,15 @@ Text inputs allow users to enter a single line of text.
...
-This checkbox has been built according to [WCAG 2.0 AA](https://www.w3.org/TR/WCAG20/) standards and all government services should strive to meet this level.  This component successfully includes the following accessibility features:
-
-### Screenreaders
-* Proper ALT labels are included
-* Language tag has been set to english
+In addition to [fundamental accessibility]() requirements, this component meets or exceeds the [WCAG 2.0 AA](https://www.w3.org/TR/WCAG20/) standards by incorporating the following criteria.
 
 ### Colour Contrast
-* Contrast ratios are at least 7:1 for normal text and 4.5:1 for large text
+* Thick boarders help better define the text input area
```
dlevineBC: "Borders with higher contrast help..., for those with visual impairments"

"Boarders" should be "borders"

components/text_input/README.md
```diff
@@ -37,19 +37,15 @@ Text inputs allow users to enter a single line of text.
...
 
 ### Keyboard Navigation
-* Supports tab navigation for use without a mouse
+* Yellow focus supports tab navigation for use without a mouse
```
dlevineBC: The focus state in general supports the tab navigation. Better for the rational to include why the custom yellow focus is used. This is also a focus state that will be used across the components so it could just be included in the main accessibility page

components/primary_button/README.md
orinevares: @dlevineBC I put in the description just for normal text since that should be the only text in a button. This is specifically stating that the text retains sufficient colour contrast when the hover state kicks in.

PR #100, opened 2019-02-06
Embedding Live Code Snippets

orinevares:
- Adding additional HTML file keep code in separate files
- Adding CSS inline to sample.html files
- Adding inline code snippet to MD files


sheaphillips:
Some of the previews aren't being displayed as expected. See attached screenshots.  Some of them  may need dev work on the componentpreview thingy, specifically around sizing of the iframe. I will take a look at these.

Also, there is an inconsistency with the class names across sample files/css that should be rationalized.  e.g. `.BC-Gov-PrimaryButton {..}` in components/primary_button.css vs `.checkbox` in components/checkbox/style.css`. We should discuss which approach (using bc-gov or similar in the class names vs. not) we want to take.  We can resolve this later.

screenshot 2019-02-08 11 44 07
screenshot 2019-02-08 11 44 29
screenshot 2019-02-08 11 44 36
screenshot 2019-02-08 11 44 44
screenshot 2019-02-08 11 45 20
screenshot 2019-02-08 11 45 28
screenshot 2019-02-08 11 45 33
screenshot 2019-02-08 11 45 38

components/primary_button/sample.html
```diff
@@ -5,6 +5,26 @@
     <link rel="stylesheet" href="./style.css">
     <link href="https://fonts.googleapis.com/css?family=Noto+Sans" rel="stylesheet">
     <title>Primary Button</title>
+    <style>
+      .BC-Gov-PrimaryButton {
```
sheaphillips: We should come up with a convention around class names and determine whether or not we want to use `bc-gov` or similar in sample classnames.  There pros and cons to each approach.  I've opened a couple of issues related to this.

PR #106, opened 2019-02-11
iframe updates, recommended badge, and last updated formatting

orinevares:


orinevares:
@sheaphillips  thanks! Starting the edits now.

orinevares:
@sheaphillips going to merge now. The URL is still not working for the logos in the header and navbar. Patrick tried a few tricks but it will require more time. Created a new issue in the icebox for now.

PR #108, opened 2019-02-11
Add Beta Indicator page

dlevineBC:
This is just a draft version. We are blocked by not having the HTML and CSS code


orinevares:
Looks good without the specific meaning around what Beta means and if it requires feedback. That can be looped in once we decide the meaning behind BETA.

As for the HTML and CSS code its just the checkbox code right now. Perhaps delete the code and just have the files ready for the correct code?

dlevineBC:
Okay I've deleted the code for the checkbox and left it empty, so it will be ready for when we can put in the correct code

orinevares:
Ok looks good to merge!

PR #115, opened 2019-02-15
Filling out accessibility info on component pages

orinevares:


components/checkbox/README.md
```diff
@@ -40,15 +40,20 @@ Based on [research](https://designnotes.blog.gov.uk/2016/11/30/weve-updated-the-
...
+
+> > "This is an unselected checkbox, checkbox not checked."
 
-### Keyboard Navigation
-* Supports tab navigation for use without a mouse
+### Target Areas
+* Checkbox size is larger than default settings to provide a larger target area
+
+### Colour Contrast
+* When checked, the checkmark exceeds a [contrast ratio](https://webaim.org/articles/contrast/) of at least 7:1 with the blue checkbox fill
```
dlevineBC: This isn't just for checked because the unchecked order also has the same contrast ratio

PR #165, opened 2019-03-25
Png icon edits

orinevares:


orinevares:
@dlevineBC pngs have been changed to 40px min width or height

PR #167, opened 2019-03-26
Callout

orinevares:


components/callout/callout.md
```diff
@@ -0,0 +1,69 @@
...
+
+## Don’t Use this For
+* Warning notifications
+* Page alerts
+
+## Design Guidance
+* Use callouts very sparingly - it’s less effective if it’s overused.
+
+## CMS Lite Users
+Learn how to insert a callout component on your page
```
dlevineBC: Need to include the link to the CMS lite manual. The link is found in the word doc

components/callout/callout.md
```diff
@@ -0,0 +1,69 @@
...
+
+## Accessibility
+
+### Colour Contrast
+* Text on the gray background meets contrast requirements
+* The thick left border conveys information using more than just colour
+
+### ARIA
+* Include the following ARIA tagging so that screenreaders read this content with emphasis:
+role="alert" aria-live="assertive"
```
dlevineBC: Include code tags so it reads this portion as code

components/callout/callout.md
```diff
@@ -0,0 +1,69 @@
...
+* This code calls attention to the content, and is for information that is important enough to be announced to the user as soon as possible, but it is not necessary to immediately interrupt the user.
```
dlevineBC: Edit, rather than say "code" let's be more specific and say "This ARIA tagging..."

components/callout/callout.md
```diff
@@ -0,0 +1,69 @@
...
+### ARIA
+* Include the following ARIA tagging so that screenreaders read this content with emphasis:
+role="alert" aria-live="assertive"
+* This code calls attention to the content, and is for information that is important enough to be announced to the user as soon as possible, but it is not necessary to immediately interrupt the user. 
+
+## Code
+### HTML
+
+```html
+<div class=”bcgov-callout”>
```
dlevineBC: Needs the ARIA tagging

components/callout/sample.html
```diff
@@ -0,0 +1,30 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <link href="https://fonts.googleapis.com/css?family=Noto+Sans" rel="stylesheet">
```
dlevineBC: We can't use the google api. This needs to be removed

components/callout/callout.md
```diff
@@ -0,0 +1,69 @@
...
+<p>Text here</p>
+</div>
+```
+
+### CSS
+
+```css
+.bcgov-callout {
+Padding: 25px;
```
dlevineBC: minor detail, but change the css properties to lowercase

components/callout/callout.md
```diff
@@ -0,0 +1,69 @@
...
+## CMS Lite Users
+Learn how to [insert a callout component](https://www2.gov.bc.ca/gov/content/governments/services-for-government/policies-procedures/web-content-development-guides/cms-lite-manual/ui-guidelines/text-editor/callout-boxes) on your page
+
+## Accessibility
+
+### Colour Contrast
+* Text on the gray background meets contrast requirements
+* The thick left border conveys information using more than just colour
+
+### ARIA
```
dlevineBC: Let's remove the ARIA until we get some clarification on this. I don't think this is actually necessary for this component.

components/callout/callout.md
```diff
@@ -0,0 +1,69 @@
...
+* Include the following ARIA tagging so that screenreaders read this content with emphasis:
+```role="alert" aria-live="assertive"```
+* This ARIA labeling calls attention to the content, and is for information that is important enough to be announced to the user as soon as possible, but it is not necessary to immediately interrupt the user. 
+
+## Code
+### HTML
+
+```html
+<div class=”bcgov-callout” role="alert" aria-live="assertive">
+<p>Text here</p>
```
dlevineBC: Let's use a real example. I grabbed this from MSP: 

By March 1, 2018, eligible residents of B.C. are expected to renew enrolment in the Medical Services Plan (MSP) and get a BC Services Card.

components/callout/callout.md
dlevineBC: Remove as well from the code examples as well

PR #173, opened 2019-03-28
Secondary button

orinevares:


components/secondary_button/README.md
```diff
@@ -0,0 +1,88 @@
+---
+description: Secondary buttons allow users to carry out a supporting task
```
dlevineBC: This description should match the description language below. Here you say task, but below you say action. Action is probably the better term

components/secondary_button/README.md
```diff
@@ -0,0 +1,88 @@
...
+    font-weight: 700;
+    letter-spacing: 1px;
+    cursor: pointer;
+    color: #003366;
+    }
+.BC-Gov-SecondaryButton:hover {
+    opacity: 0.80;
+    text-decoration: underline;
+    background-color: #003366;
+    color: white;
```
dlevineBC: #FFFFFF

components/secondary_button/style.css
```diff
@@ -0,0 +1,27 @@
+.BC-Gov-SecondaryButton {
```
dlevineBC: This should probably be confirmed with a developer, but I think you should specify "background: none"; I think if you don't there's a chance that a default styling or other CSS styling could apply a background to it.

components/secondary_button/style.css
```diff
@@ -0,0 +1,27 @@
...
+    font-weight: 700;
+    letter-spacing: 1px;
+    cursor: pointer;
+    color: #003366;
+    }
+.BC-Gov-SecondaryButton:hover {
+    opacity: 0.80;
+    text-decoration: underline;
+    background-color: #003366;
+    color: white;
```
dlevineBC: #FFFFFF

components/secondary_button/style.css
```diff
@@ -0,0 +1,27 @@
+.BC-Gov-SecondaryButton {
+    border-radius: 4px;
+    border: 2px solid #003366;
+    padding: 10px 30px;
+    text-align: center;
+    text-decoration: none;
+    display: block;
+    font-size: 18px;
+    font-family: 'Noto Sans','Calibri', 'Arial', 'Sans Serif';
```
dlevineBC: Font-family: ‘Noto Sans’, Verdana, Arial, sans-serif;

components/secondary_button/sample.html
```diff
@@ -0,0 +1,41 @@
...
+    <style>
+    .BC-Gov-SecondaryButton {
+      border-radius: 4px;
+      border: 2px solid #003366;
+      padding: 10px 30px;
+      text-align: center;
+      text-decoration: none;
+      display: block;
+      font-size: 18px;
+      font-family: 'Noto Sans','Calibri', 'Arial', 'Sans Serif';
```
dlevineBC: Font-family: ‘Noto Sans’, Verdana, Arial, sans-serif;

components/secondary_button/README.md
```diff
@@ -0,0 +1,88 @@
...
+
+## Use This For
+* A supporting action you want users to take on your site.
+
+## Don't Use This For
+*	Primary actions on your page such as 'Submit', or 'Start'. Use a [Primary Button](https://developer.gov.bc.ca/Design-System/Primary-Button) instead.
+
+## Design Guidance
+*	Use for supporting actions or options you don't want users to default to
+*	Button labels should be as short as possible with action words that clearly explain what will happen when the button is clicked (eg. Back, Cancel, or Don't Agree)
```
dlevineBC: "Don't Agree" doesn't really explain a clear action. Would suggest removing this example

components/secondary_button/README.md
```diff
@@ -0,0 +1,88 @@
...
+* Do not use images for buttons
+
+## Rationale
+The rounded edges on the button help differentiate it from other layout items on a page. The underline on hover ensures colour is not the only visual means of conveying information about the button being clickable. The transparent background helps the button reduce its visual prominance amongst other elements and primary buttons on a page.
+
+## Behaviour
+1. Button label should appear underlined on hover
+2. Button text and border colour should revert to #003366 when in the active state
```
dlevineBC: It seems that "label" and "text" are used interchangeably here. When you say active state, do you mean the default state?

components/secondary_button/README.md
```diff
@@ -0,0 +1,88 @@
...
+## Accessibility
+In addition to [fundamental accessibility]() requirements, this component meets or exceeds the [WCAG 2.0 AA](https://www.w3.org/TR/WCAG20/) standards by incorporating the following criteria.
+
+### Screenreaders
+As read using ChromeVox
+
+> *"Cancel, button."*
+
+### Colour Contrast
+* [Contrast ratio](https://webaim.org/resources/contrastchecker/) exceed 7:1 for text on button when in static and hover states.
```
dlevineBC: This isn't really accurate because the background behind these buttons could be any colour, but the primary blue holds up pretty well. Maybe add "...when on a white or light grey background"

components/secondary_button/README.md
orinevares: the active state is when the user clicks on the button. The sequence would go default->hover>(click)>active while clicking.

components/secondary_button/README.md
```diff
@@ -0,0 +1,90 @@
...
+<component-preview path="components/secondary_button/sample.html" height="100px" width="800px"> </component-preview>
+
+## Use This For
+* A supporting action you want users to take on your site.
+
+## Don't Use This For
+*	Primary actions on your page such as 'Submit', or 'Start'. Use a [Primary Button](https://developer.gov.bc.ca/Design-System/Primary-Button) instead.
+
+## Design Guidance
+*	Use for supporting actions or options you don't want users to default to
```
dlevineBC: I reviewed this line with Kaitlyn and we determined we can just get rid of it. This is information that's duplicated in the "use this for" and the "don't use this for"

components/secondary_button/README.md
```diff
@@ -0,0 +1,90 @@
...
+
+# Secondary Button
+Secondary buttons allow users to carry out a supporting action on your service, such as Back or Cancel.
+
+## Visual Example
+
+<component-preview path="components/secondary_button/sample.html" height="100px" width="800px"> </component-preview>
+
+## Use This For
+* A supporting action you want users to take on your site.
```
dlevineBC: These could be plural as there could be a few secondary actions, so maybe say "Supporting actions user can take on your site"

PR #230, opened 2019-11-20
Included BCSans in CSS font-family

lclarko:
I've added "BCSans" to the font-family CSS, within the primary button component readme. This was originally requested by @davidcusack via a devhub content issue.

### DIFF
--- a/components/primary_button/README.md
+++ b/components/primary_button/README.md
@@ -70,7 +70,7 @@ As read using ChromeVox
     text-decoration: none;
     display: block;
     font-size: 18px;
-    font-family: ‘Noto Sans’, Verdana, Arial, sans-serif;
+    font-family: BCSans, ‘Noto Sans’, Verdana, Arial, sans-serif;
     font-weight: 700;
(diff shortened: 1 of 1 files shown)


components/primary_button/README.md
```diff
@@ -70,7 +70,7 @@ As read using ChromeVox
     text-decoration: none;
     display: block;
     font-size: 18px;
-    font-family: ‘Noto Sans’, Verdana, Arial, sans-serif;
+    font-family: BCSans, ‘Noto Sans’, Verdana, Arial, sans-serif;
```
dlevineBC: I think we need quotes around BCSans

davidcusack: You are right. Sorry I missed that

lclarko: My mistake - I've wrapped BCSans in quotes, as it is a non-generic font-family/required by CSS

lclarko: @davidcusack  You had it right in the issue - I omitted the quotes, in error

PR #236, opened 2020-03-09
Update prototyping_tools.md

dlevineBC:
Adding in download for XD file

### DIFF
--- a/components/about/prototyping_tools.md
+++ b/components/about/prototyping_tools.md
@@ -6,5 +6,8 @@ author: orinevares
 
 # Prototype with the Design System
 
-## Sketch App (Coming Soon)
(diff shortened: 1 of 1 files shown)


ty2k:
@dlevineBC, it looks like the `download` attribute in the link tag gets stripped out automatically by GitHub's Markdown parser. I've tried the following suggestions from Stack Overflow on [my fork](https://github.com/ty2k/design-system/tree/dlevineBC-xd) to see if I could get that functionality working:

- `id="raw-url"` in the `<a>` tag, like what GitHub uses on their Download buttons for files
- `download` and `target="_blank"` in the `<a>` tag
- `type="application/octet-stream"` in the `<a>` tag

All of those get stripped out by the parser, likely by design.

I will merge this as the `download` attribute does no harm and shows the intent.

PR #260, opened 2020-09-08
Linting, apostrophe, and comment fixes in Typography code samples

### LINKED ISSUE #257: Devhub Issue: Typography CSS Comments & `‘` issue
> path: (do not delete) https://github.com/bcgov/design-system/blob/master/styles/typography/typography.md
 > (do not delete) devhub page: /Typography

## Devhub Content Issue

Hi! 

- The comments in the `@font-face` section may be incorrect. Believe it should be `/* Optimized for very modern browsers */ ` instead of `/ Optimized for very modern browsers / `
- The code in the `Font Stack` section was causing problems with my babel/webpack setup. It could be causing problems for others. Had to change `‘` to `'`, eg: `font-family: ‘BCSans’, ‘Noto Sans’, Verdana, Arial, sans-serif;` -> `font-family: 'BCSans', 'Noto Sans', Verdana, Arial, sans-serif;`

ty2k:
Fixes #257.

### DIFF
--- a/styles/typography/header-sample.html
+++ b/styles/typography/header-sample.html
@@ -6,30 +6,30 @@
     <link href="https://fonts.googleapis.com/css?family=Noto+Sans: 400, 700" rel="stylesheet">
     <title>Headings</title>
     <style>
-    body {
-    font-family: ‘BCSans’, ‘Noto Sans’, Verdana, Arial, sans-serif;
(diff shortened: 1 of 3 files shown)


lclarko:
I now see that &apos; is not a valid HTML 4 entity!

PR #265, opened 2021-01-29
Content update: Telephone link guidance

ty2k:
This pull requests adds new guidance to the [Links](https://developer.gov.bc.ca/Design-System/Links) page within the Design System for telephone links (`<a href="tel: ... ">`). Written guidance and code samples are included.

Sources used for composing this new documentation:
- [MDN: Linking to Telephone Numbers](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#linking_to_telephone_numbers)
- [Google Developers Web Fundamentals: Click to Call](https://developers.google.com/web/fundamentals/native-hardware/click-to-call)

Additional changes in this pull request:
- Code samples are updated for correctness and spacing/linting
- Screen reader samples are updated for correctness
- On-page component preview iframe has its height increased to accommodate new example
- "Users" language changed to "people"

### DIFF
--- a/components/link/README.md
+++ b/components/link/README.md
@@ -24,10 +24,10 @@ Links lead people to a different page.
 * Links should be describe where it leads to
 * Do not put links in words such as "Link", "Here", or "Click Here". [Learn more about descriptive links](https://accessibility.oregonstate.edu/descriptivelinks)
 * Telephone links should be used to allow people to click a phone number to make a call
-* Links should not open a new tab or window. If so, they need to have the relevant icon to give the person advanced warning. Example below:
+* Links should not open a new tab or window unless it is preferable from an accessibility perspective. Ex. opening a new tab or window allows people to remain logged in on a secured site or avoid progress lost while filling in a form. If a link opens a new tab or window, it needs to have a warning included in the text such as "(opens in new window)". Example:
 
 ```html
-<a href="knitting.html" target="_blank">Application Portal (opens in new window)</a>
+<a href="/portal" target="_blank">Application Portal (opens in new window)</a>
 ```
 
 ## Behaviour
--- a/components/link/README.md
+++ b/components/link/README.md
@@ -6,15 +6,15 @@ author: orinevares
 ---
 
 ![Status](https://img.shields.io/badge/Recommended-Draft-orange.svg)
-> Last Updated: February 11, 2019
+> Last Updated: January 29, 2021
 
 # Links
 
-Links lead users to a different page.
+Links lead people to a different page.
 
 ## Example
 
-<component-preview path="components/link/sample.html" height="150px" width="800px"> </component-preview>
+<component-preview path="components/link/sample.html" height="200px" width="800px"> </component-preview>
 
 ## Don't Use This when:
 * Users need to take an action on a page such as log-in or submit, use a [primary button](../primary_button/README.md) instead.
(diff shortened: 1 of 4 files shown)


knguyenBC (commented):
- Change sentence: "Users need to take action on a page..." to "People need to take action on a page..."
- Should we not be referencing our own guide to descriptive links such as: https://www2.gov.bc.ca/gov/content/governments/services-for-government/policies-procedures/web-content-development-guides/writing-for-the-web/web-style-guide/links 
- Missing comma: "If so, they need to have the relevant..." 
- Is the 'external link' icon the same as the 'opens in new window' icon?
- The same guidance as linked above states our telephone format is 1-555-555-5555.

ty2k:
Thanks for the feedback @knguyenBC!

- For the "users" vs "people" change, this update has already been made in this pull request, but the [DevHub Links page](https://developer.gov.bc.ca/Links) is not updated to reflect this. Unfortunately there is no easy way to show you these changes on a deployed site. The best way to see the changes being made is using the [files tab](https://github.com/bcgov/design-system/pull/265/files) on this pull request. The red highlighted blocks of text represent code being removed, and the green is code being added. If this doesn't work for you, let me know Tuesday and we can share my screen to see what it all looks like in context from my machine.

- Good call on using BC Gov content for the descriptive links guidance. I've updated this pull request to use the page you suggested using its permalink. (5ae6e60)

- I've updated this pull request to fix the missing comma typo. (3a2d7be)

- For the external link icon, the page is not clear at the moment. It reads: "Links should not open a new tab or window. If so they need to have the relevant icon to give the user advanced warning. Example below:". However, the code sample that follows uses a text hint ("opens in new window") and does not use an icon. Could you please advise whether I should add an icon to this code sample, or whether I should remove the icon reference?

- I couldn't find the reference to the number format "1-555-555-5555" in the Web Style Guide. I see the [Numbers page in the Web Style Guide](https://www2.gov.bc.ca/gov/content?id=2C00DB87CE6642C88C526DD633F60E71) uses both an example with a leading "1-" country code (for a toll free number) and one without. My feeling on this is that the link text should be left up to the content creator (with leading "1-" for toll free numbers and the like, and ten digits separated by hyphens for standard in-province area codes), but the link `href` should use the international dialing format (leading `+1`) to ensure the link works when clicked anywhere in the world.

knguyenBC:
- The external link icon is the same as the open in new window icon it seems. That can be confusing. WCAG 2.0 provides two examples of how to deal with opening in a new tab/window: https://www.w3.org/TR/WCAG20-TECHS/G201.html. Let's go with example 1.  Sentence should say something like "Links should not open a new tab or window unless it is preferable from an accessibility perspective. Ex. opening a new tab or window allows people to remain logged in on a secured site or avoids progress lost while filling a form. If a link opens a new tab or window, it needs to have a warning included in the text such as Link (opens in new window). Example below:"

- Yes, please follow the numbers page in the web style guide. What I meant was that blocks of digits (country code, area code, etc) should be separated by dashes. i.e. +1 (555) 555-555 does NOT follow our web guidelines.

ty2k:
Thanks @knguyenBC, I've updated the external link guidance (17926b4) and confirmed that we are following the Web Style Guide's phone number style (dashes, not parentheses). Can you please let me know if this is good to merge?

knguyenBC:
Good to merge

PR #285, opened 2024-02-14
Docs

mkernohanbc:
This change rewrites README.md to reflect current status and future plans for the design system product. It also adds a holding message to CONTRIBUTING.md to reflect that the contribution process and criteria are likely to change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,26 @@
-# BC Government Design System [![img](https://img.shields.io/badge/Lifecycle-Maturing-007EC6)](https://github.com/bcgov/repomountie/blob/master/doc/lifecycle-badges.md)
+# B.C. Design System [![img](https://img.shields.io/badge/Lifecycle-Maturing-007EC6)](https://github.com/bcgov/repomountie/blob/master/doc/lifecycle-badges.md)
+
+> [!IMPORTANT]
+> The B.C. Design System is currently in active development. We are releasing new guidance and assets in stages. The legacy design system will eventually be fully deprecated and archived.
+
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -1,46 +1,54 @@
...
-Please note that this project is released with a [Contributor Code of Conduct](CODE_OF_CONDUCT.md). 
-By participating in this project you agree to abide by its terms.
+## Legacy documentation
+> [!WARNING]
+> The previous B.C. government design system is not actively supported. These resources are provided for reference only. These resources will be officially deprecated and archived after the new BCDS component library is released.
+
+Documentation and components for the legacy design system are accessible in this repo and on DevHub:
+
+* [Developer documentation](https://developer.gov.bc.ca/Design-System/About-the-Design-System)
+* [Legacy components on DevHub](https://developer.gov.bc.ca/components)
```
ty2k, line 50: We should leave out this "Legacy components" link. It's pointing to a landing page for any document with the label "Components" in DevHub, which isn't limited to the Design System stuff. Example of how that label gets applied: https://github.com/bcgov/devhub-app-web/blob/master/web/topicRegistry/design-system.json#L54

Better to just link to that "Developer documentation" page above. When you include the `/Design-System` part of the path segment as you have done, you get the full left navigation with the list of Design System components there.

ty2k (approved):
Looks great! One small change suggestion from me, but not a blocker if you want to merge as-is.

README.md
mkernohanbc: Good catch — I've just removed that link.

PR #299, opened 2024-03-13
Update Storybook branding

mkernohanbc:
This PR contains an initial branding pass on our Storybook instance:

- 72819c8: implements a custom Storybook theme (`BCDS_theme`) with B.C. gov branding and colour scheme
- 04bd601 and f6f7c4c: adds BC Sans and B.C. Design Tokens packages as dependencies

The remaining commits reskin the Storybook UI to use BCDS colour scheme and layout (borders, corners, etc.)

Tokens implementation will need to be refactored to use 3.0.0 tokens.

PR #300, opened 2024-03-13
Fix logo in Storybook deployment

ty2k:
Currently, the logo breaks in the Storybook layout because the directory traversal in the `` tag's `src` attribute isn't working:

Storybook in OpenShift with logo broken

 To fix, I've added a new directory for static files that will be available to Storybook after a Dockerfile build, and placed a copy of the logo there: `storybook-public`. I've updated the layout to point to this new version of the file, and it works after a local Dockerfile build:
 
Storybook locally in a Docker container with logo working


mkernohanbc (approved):
Thanks for catching this — that logo was being intermittently janky when I was working on #299, but I thought I'd figured it out!

PR #311, opened 2024-03-21
Refactor Storybook theme to use 3.0.0-rc1 tokens

mkernohanbc:
This change updates the `bcdsTheme` implemented in #299 to use the latest version of the design tokens package (3.0.0-rc1). 

It also includes a small tweak to the Storybook UI layout (72d205b), setting the control panel in story view to appear at the bottom of the screen, rather than as a sidebar on the right-hand side.


ty2k (approved):
All the new changes work great for the Storybook theming! 🚀 

As discussed, we will work in this `react-components-with-tokens-v3` feature branch until all the individual components have been updated to use the v3 tokens, and then we can merge that feature branch back into `main`.

PR #335, opened 2024-04-22
Add deprecation notice to legacy 'about' page

mkernohanbc:
This change makes a small edit to [the legacy 'About the design system' page](https://classic.developer.gov.bc.ca/Design-System/About-the-Design-System) to note that it is archived content and direct users to the new content hub on gov.bc.ca.

This change should not be merged until after the new DS site is published.

### DIFF
--- a/components/about/about.md
+++ b/components/about/about.md
@@ -4,7 +4,8 @@ title: About the Design System
 author: ksingbeil
 ---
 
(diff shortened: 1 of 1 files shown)


ty2k (approved):
This looks good to me, but I'm unable to build Classic DevHub locally to actually have the app try and render the GitHub `[!WARNING]` block. DevHub uses `remark` for Markdown parsing, and I have no idea what a GitHub-specific tag like that will do to it. Classic DevHub is also no longer maintained, and this change won't make it to the deployed site unless the Developer Experience team runs a new build of it for some reason. I think it's safe enough to merge this when you like.

PR #339, opened 2024-04-30
@bcgov/design-tokens v3.0.0-rc2

### LINKED ISSUE #312: Add CJS as an export format for design tokens
Context:

- Configuring our Storybook instance to use the bcgov/design-tokens package for custom styling
- Attempting to use [variables.js](https://github.com/bcgov/design-system/blob/main/packages/design-tokens/build/js/variables.js) to implement design tokens in [main.ts](https://github.com/bcgov/design-system/blob/main/packages/react-components/.storybook/main.ts) caused compilation errors and build failure
- Reimplementing tokens in CJS resolved issue and allowed Storybook build to complete

This suggests that our current ECMAScript-only approach to shipping JS variables may not be suitable for all use cases. Let's investigate whether it's feasible/straightforward it is to add CJS to our Tokens Studio build pipeline.

### LINKED ISSUE #321: Provide TypeScript definitions for @bcgov/design-tokens
### Related component

Design tokens npm package

### Description

The `@bcgov/design-tokens` npm package currently doesn't ship TypeScript definitions for the tokens and it should.

### Rationale

Not shipping type definitions can cause extra work for TypeScript projects using the file: the compiler might need to know the type of the token being imported, so the developers will have to either write their own definition module or cast the imported token value. Both cause extra work for the end developer when this might be something we can generate and supply automatically as part of the Style Dictionary/sd-transforms pipeline.

### Additional information

See example: https://github.com/ratheesh-aot/met-public/commit/7b2147f620574b00bba9ebfb1389252956630d9b

ty2k:
This PR updates the design tokens package to add CommonJS variables (closes #312) and TypeScript definitions (closes #321).

CommonJS variables can be imported like:
```js
const tokens = require("@bcgov/design-tokens/cjs");
```

One new breaking change is included in 181da663cd65353b652a820729f569b5e187f977: The output filenames for JavaScript build targets change from `variables.js` to `index.js`. This is to shorten the import statements in code that consumes the tokens package.

Before:
```js
import * as tokens from "@bcgov/design-tokens/js/variables.js";
```

After:
```js
import * as tokens from "@bcgov/design-tokens/js";
```


mkernohanbc (approved):

ty2k:
Published `@bcgov/design-tokens` @ v3.0.0-rc2: https://www.npmjs.com/package/@bcgov/design-tokens/v/3.0.0-rc2

PR #346, opened 2024-05-06
remove bad styling from Storybook theme

mkernohanbc:
This change removes two custom styling rules from the `bcdsTheme`, because they break part of the Storybook UI by making buttons in the 'Controls' sidebar of story view impossible to read.


ty2k (approved):
Looks good to me!

Story screen showing updated button styling

Issue #354, opened 2024-05-21
Select - fix colour contrast in secondary label


mkernohanbc:
The secondary label "(optional)" in the select component currently fails WCAG AA colour contrast requirement. The required contrast ratio is 4.5:1.

This label is currently styled using `--typography-color-disabled`. It should more properly use `--typography-color-secondary`, which should pass the contrast ratio requirement.

Axe test result in Storybook:
![Image](https://github.com/bcgov/design-system/assets/135075821/affd2f3c-e0cf-4d81-88d7-9badccb406c8)

PR #355, opened 2024-05-21
Select - change colour of "(optional)" label

mkernohanbc:
This PR changes the colour of a secondary label in the select component to use the context-appropriate token `--typography-color-secondary` instead of `--typography-color-disabled`.

This is should resolve the accessibility issue described in #354


ty2k:
Looking good!

Select component in Storybook with updated optional label color

ty2k (approved):
🚀

PR #356, opened 2024-05-23
GitHub commit link in Storybook build

ty2k:
This PR uses the [`GITHUB_SHA` workflow environment variable](https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables) to display a link within the React component library Storybook that points back to the commit that kicked off the Storybook build. This allows visitors to our Storybook instances to understand which version of the React component library code the Storybook documentation refers to. The version of the package from `/package/react-components/package.json` is also shown.

Example Storybook screenshot when the GitHub SHA variable is available:
Storybook after a build

Example Storybook screenshot when the GitHub SHA isn't present, like in local development:
Storybook in local development

### DIFF
--- a/.github/workflows/build_react_component_library_apps.yaml
+++ b/.github/workflows/build_react_component_library_apps.yaml
@@ -27,7 +27,7 @@ jobs:
           driver-opts: network=host
 
       - name: Build image with docker build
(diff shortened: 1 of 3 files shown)


mkernohanbc (approved):
Looks good to me! Behaves as expected when I build Storybook locally:

Screenshot 2024-05-23 at 16 14 25

PR #362, opened 2024-05-28
Add missing typography-color-secondary to Select label

ty2k:
Following #354 , this adds a missing `color` to the Select component's label. Without this PR, the browser default black is being applied to the label:

Storybook from OpenShift dev environment

After this PR:
Storybook from local dev environment

This makes the "main" label have the same color as the "(optional)" tag when it is present. Note that we are currently missing a canonical example of what this label should actually look like with the "(optional)" tag in our pre-release Figma component file:

Figma Dropdown example from May 28

### DIFF
--- a/packages/react-components/src/components/Select/Select.css
+++ b/packages/react-components/src/components/Select/Select.css
@@ -9,6 +9,7 @@
 
 /* Label above select input */
 .bcds-react-aria-Select--Label {
+  color: var(--typography-color-secondary);
   font: var(--typography-regular-small-body);
 }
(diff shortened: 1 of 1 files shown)


Philip-Cheung (approved):

mkernohanbc (commented):
My interpretation is that the main label should use `--typography-color-primary`, and the "(optional)" label should use `--typography-color-secondary`, but I think that @Philip-Cheung has the decision on what's correct here — adding him as a reviewer.

(Philip, this relates to the Figma change requested in https://github.com/orgs/bcgov/projects/106/views/3?pane=issue&itemId=63792131)

Issue #367, opened 2024-05-30
Select/dropdown - suppress known false positive in Storybook accessibility tests


mkernohanbc:
The automated accessibility testing we run within Storybook (using the [storybook-addon-a11y](https://storybook.js.org/addons/@storybook/addon-a11y) extension) throws a 'serious violation' warning on the Select component: 

> ARIA hidden element must not be focusable or contain focusable elements

![Image](https://github.com/bcgov/design-system/assets/135075821/be9f71a9-82ac-49ae-88e0-dc7848b98405)

This is [a known false positive within RAC's Select component](https://react-spectrum.adobe.com/react-aria/Select.html#false-positives): 

> Select may trigger a [known accessibility false positive](https://github.com/adobe/react-spectrum/wiki/Known-accessibility-false-positives#picker) from automated accessibility testing tools. This is because we include a visually hidden select element alongside the Select to specifically aid with browser form autocomplete and hide it from screen readers via aria-hidden since users don't need to interact with the hidden select. We manage focus internally so that focusing this hidden select will always shift focus to the Select's trigger button instead. Automated accessibility testing tools have no way of knowing that we manage the focus in this way and thus throw this false positive.

RAC provides an attribute we can target to override and suppress this warning in storybook-addon-a11y.

mkernohanbc:
Thinking about this a little more, probably a good idea to include a note about this false positive in the docs for the component too — it's going to pop up if users are running similar automated tests, they need to know what's going on. @ty2k what do you think?

PR #368, opened 2024-05-30
override automated test for aria-hidden-focus in Select component

mkernohanbc:
This change overrides automated accessibility testing for `aria-hidden-focus` in the Select component in Storybook, in order to suppress a known false positive (see #367) in the [storybook-addon-a11y](https://storybook.js.org/addons/@storybook/addon-a11y) reporting.

False positive violation no longer shown in accessibility panel:

Screenshot 2024-05-30 at 11 07 02

### DIFF
--- a/packages/react-components/src/stories/Select.stories.tsx
+++ b/packages/react-components/src/stories/Select.stories.tsx
@@ -7,6 +7,13 @@ const meta = {
   component: Select,
   parameters: {
     layout: "centered",
+    a11y: {
+      config: {
+        // overrides automated accessibility testing for aria-hidden-focus to suppress a known false positive
+        // see https://react-spectrum.adobe.com/react-aria/Select.html#false-positives for more information
+        rules: [{ id: 'aria-hidden-focus', enabled: false }],
+      }
+    }
   },
   argTypes: {
     size: {


mkernohanbc:
Thinking about this a little more, probably a good idea to include a note about this false positive in the docs for the component too — it's going to pop up if users are running similar automated tests, they need to know what's going on. @ty2k what do you think?

ty2k (approved):
I think the combination of disabling the `aria-hidden-focus` rule here in the Select stories file and also documenting it (done in #378) is a great solution for this issue. Because this is a genuine false-positive and not an a11y issue that a real user will experience, disabling the rule is fine.

Since this PR was opened, we added the `test-storybook` script using `@storybook/test-runner` and `axe-playwright` in #383. With that script, the `aria-hidden-focus` error isn't currently being triggered. If that were to change, we could alter the `test-runner.ts` configuration to work around it, as described in the [@storybook/test-runner readme here](https://github.com/storybookjs/test-runner?tab=readme-ov-file#accessibility-testing).

Here's what it error looks like in the Storybook front-end prior to the change in this PR, for posterity: 
Select component in Storybook showing a11y error

Issue #388, opened 2024-06-20
Select/dropdown - refine error handling


mkernohanbc:
Our current implementation (in both Figma and React) indicates an error/invalid state by turning the selector border red:

- Figma: https://www.figma.com/design/Fr33ALEYGvM66CSw2iwxzb/Components?node-id=2518-2519&t=uLivxfoEJXFWKOXO-4
- React: https://design-system-react-components-storybook-ed91fb-dev.apps.silver.devops.gov.bc.ca/?path=/story/components-select-select--invalid

This [isn't sufficient to pass WCAG SC 1.4.1](https://www.w3.org/WAI/WCAG21/Techniques/failures/F81) — we can't _only_ use colour to convey information. 

RAC Select [appears to have built-in support for displaying an error message](https://react-spectrum.adobe.com/react-aria/Select.html#fielderror), but we don't currently appear to be leveraging this.

An invalid selection should be indicated by both:

1. A red border around the selector
2. An error message displayed below the selector

Let's do another pass on this, please: 

- [x] Add error message support to Figma component
- [x] Add error message support to React component

**Update**: [New design in Figma](https://www.figma.com/design/2OhHsH2ztSFUwDcVzxug7Z/BC-Design-System?node-id=3572-4278&t=iltwjpSV8xQQThK4-4)

PR #390, opened 2024-06-20
Select/dropdown - improve error handling and focus state

mkernohanbc:
This PR contains minor tweaks to the Select component, to address the issues documented in #388 (based on work done for #389):

- 1c0c452: shows a visible error icon to the Select `<Button>` element when the `isInvalid` prop is passed
- a56cd53: first pass at implementing validation and error handling using `FieldError` and `ValidationResult`. Creates an error message slot underneath the Select `<Button>`

### DIFF
--- a/packages/react-components/src/components/Select/Select.css
+++ b/packages/react-components/src/components/Select/Select.css
@@ -17,12 +17,6 @@
   color: var(--typography-color-disabled);
 }
 
-/* "(optional)" text after Label*/
-.bcds-react-aria-Select--Label > .optional {
-  padding: var(--layout-padding-none) var(--layout-padding-xsmall);
-  color: var(--typography-color-secondary);
-}
-
 /* Error message */
 .bcds-react-aria-Select--Error {
   font: var(--typography-regular-small-body);
--- a/packages/react-components/src/components/Select/Select.css
+++ b/packages/react-components/src/components/Select/Select.css
@@ -9,6 +9,8 @@
 
 /* Label above select input */
 .bcds-react-aria-Select--Label {
+  display: flex;
+  padding: var(--layout-padding-small) var(--layout-padding-none);
   color: var(--typography-color-secondary);
   font: var(--typography-regular-small-body);
 }
@@ -18,6 +20,7 @@
 
 /* "(optional)" text after Label*/
 .bcds-react-aria-Select--Label > .optional {
+  padding: var(--layout-padding-none) var(--layout-padding-xsmall);
   color: var(--typography-color-secondary);
 }
 
--- a/packages/react-components/src/components/Select/Select.css
+++ b/packages/react-components/src/components/Select/Select.css
@@ -21,6 +21,12 @@
   color: var(--typography-color-secondary);
 }
 
+/* Error message */
+.bcds-react-aria-Select--Error {
+  font: var(--typography-regular-body);
+  color: var(--typography-color-danger);
+}
+
 /* Select input equivalent */
 .bcds-react-aria-Select--Button {
   background-color: var(--surface-color-forms-default);
--- a/packages/react-components/src/components/Select/Select.css
+++ b/packages/react-components/src/components/Select/Select.css
@@ -26,6 +26,8 @@
 
 /* Error message */
 .bcds-react-aria-Select--Error {
+  display: flex;
+  padding: var(--layout-padding-small) var(--layout-padding-none);
   font: var(--typography-regular-small-body);
   color: var(--typography-color-danger);
 }
--- a/packages/react-components/src/components/Select/Select.css
+++ b/packages/react-components/src/components/Select/Select.css
@@ -8,10 +8,9 @@
 
 /* Label above select input */
 .bcds-react-aria-Select--Label {
-  display: flex;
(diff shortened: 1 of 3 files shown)


mkernohanbc:
Leaving this in draft for the moment, currently having trouble validating whether the `errorMessage` slot is behaving as expected — need to do some more digging on how to put the component into different error/validation failure states to confirm that this is working.

mkernohanbc:
Folding some additional tweaks into this PR:

- [45d643e](https://github.com/bcgov/design-system/pull/390/commits/45d643ee77599ba87be52b9259f2e2981991c276): implements the new focus state style, closes #394 
- [541ad99](https://github.com/bcgov/design-system/pull/390/commits/541ad99da180a3ed4fecc620f4edbeae65733e87): adds margin to select `<button>`, to adjust the spacing between the label and the selector

mkernohanbc:
[19c1b3b](https://github.com/bcgov/design-system/pull/390/commits/19c1b3b36bf6aa9938c6b3d78ae723a13e27e98a) reverts/replaces [541ad99](https://github.com/bcgov/design-system/pull/390/commits/541ad99da180a3ed4fecc620f4edbeae65733e87):

- adds correct padding to label element and sets it to `display:flex`, instead of adding margin to the button
- adds horizontal padding to the "(optional)" label `<span>` 
- changes `errorMessage` slot styling to use smallBody token

packages/react-components/src/components/Select/Select.css
```diff
@@ -9,6 +9,8 @@
 
 /* Label above select input */
 .bcds-react-aria-Select--Label {
+  display: flex;
```
ty2k, line 12: Let's revert the changes inside of `.bcds-react-aria-Select--Label` and its children. We want this label to keep its default `display: inline` behaviour. If we were to set this to `display: flex`, things break when text wraps, and the "(optional)" chunk starts floating next to the label.

Current `main`:
Existing Select with very long label

Proposed in this branch:
Screenshot 2024-07-17 at 1 16 54 PM

We originally had a `<span>` around the "(optional)" chunk because it was used to color the chunk a different shade than the main label. If we are no longer doing that, we can simplify the markup in the component code and remove the extra span, and just build a single string inside of `<Label>`. However, we don't have a current copy of the design for this component that uses the "(optional)" chunk to confirm that this is the requirement.

packages/react-components/src/components/Select/Select.tsx
```diff
@@ -123,8 +146,12 @@ export default function Select<T extends object>({
                 return "Select an item";
               }}
             />
+            {isInvalid && iconError}
```
ty2k, line 149: I like this inline icon addition to align better with the text fields. We should get the designs updated to reflect this, even though it's a rare state for a select menu.

Screenshot 2024-07-17 at 1 33 22 PM

packages/react-components/src/components/Select/Select.css
```diff
@@ -18,9 +20,18 @@
 
 /* "(optional)" text after Label*/
 .bcds-react-aria-Select--Label > .optional {
+  padding: var(--layout-padding-none) var(--layout-padding-xsmall);
   color: var(--typography-color-secondary);
 }
 
+/* Error message */
+.bcds-react-aria-Select--Error {
+  display: flex;
```
ty2k, line 29: Is the intent here to have this element hold more than just a text string? If so, `display: flex` might be warranted, but if not, we can leave this out.

packages/react-components/src/components/Select/Select.tsx
```diff
@@ -86,13 +90,32 @@ function ChevronUp() {
   );
 }
 
+/* Icon displayed when input is in invalid state */
```
ty2k, line 93: Keep this here for now and we can replace it with a common file in a new PR when more of these outstanding ones are merged and it's clear where it's needed.

packages/react-components/src/components/Select/Select.css
```diff
@@ -48,7 +59,9 @@
 }
 .bcds-react-aria-Select--Button[data-focused] {
   border-color: var(--surface-color-border-active);
-  outline: none;
+  outline: solid var(--layout-border-width-medium)
```
ty2k, line 62: This looks great! 👍 👍 

Select component focused with popover open

ty2k (changes requested):
Most excellent @mkernohanbc 🚀 . There are a couple tweaks needed in CSS, but the new error slot stuff looks awesome.

packages/react-components/src/components/Select/Select.css
mkernohanbc: Fixed in [6805b3f](https://github.com/bcgov/design-system/pull/390/commits/6805b3f753076d54b210f7a0eee3d5dd30ababe3)

packages/react-components/src/components/Select/Select.css
mkernohanbc: fixed in [f415819](https://github.com/bcgov/design-system/pull/390/commits/f415819499ddec73f9dafffa279e984238970856)

ty2k (approved):
Great stuff ✅

PR #392, opened 2024-06-20
Update button focus state to new design

mkernohanbc:
This PR changes the styling of the focus state for the button component, offsetting the outline on `[data-focused]` so that it's more visually prominent. 

New focus state is illustrated here:
Screenshot 2024-06-20 at 15 02 09
Screenshot 2024-06-20 at 15 02 33


ty2k (approved):
🚀

PR #396, opened 2024-06-25
Update tag focus and select states to new design

mkernohanbc:
This PR implements the new offset-ring focus state design on the tag component.

This change makes it easier to distinguish when a tag is in focus, selected or both:

Screenshot 2024-06-25 at 13 55 35

[New design in Figma](https://www.figma.com/design/2OhHsH2ztSFUwDcVzxug7Z/BC-Design-System?node-id=3347-129&m=dev), for reference.

### DIFF
--- a/packages/react-components/src/components/Tag/Tag.css
+++ b/packages/react-components/src/components/Tag/Tag.css
@@ -77,7 +77,7 @@
 /* Focused */
 .bcds-react-aria-Tag[data-focused] {
   outline: 2px solid var(--surface-color-border-active);
-  outline-offset: 0;
+  outline-offset: var(--layout-margin-hair);
 }
 
 /* Disabled */
--- a/packages/react-components/src/components/Tag/Tag.css
+++ b/packages/react-components/src/components/Tag/Tag.css
@@ -70,13 +70,16 @@
 
 /* Selected */
 .bcds-react-aria-Tag[data-selected] {
-  border: 2px solid var(--surface-color-border-active);
-  padding: 1px 7px;
+  border-radius: var(--layout-border-radius-small);
+  border: var(--layout-border-width-medium) solid
+    var(--surface-color-border-active);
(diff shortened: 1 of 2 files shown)


ty2k:
The new focus style looks great!

One existing issue with this component is that our current markup and styles will cause some weirdness for the "Selection" story, as shown in your screenshot. The existing markup and style for the non-selected tag is causing its text and icon to sit visually higher in the container than it should be.

Here's a zoomed in example of what's in this branch currently:
Selected TagGroup using this branch's code

I've added a [branch here](https://github.com/bcgov/design-system/tree/feature/refactor-tag-markup) where I've removed an unnecessary `<div>` from the markup and simplified the styling. This is what that version of the code looks like:
Selection TagGroup with updated markup

If you open those two screenshots in separate tabs and switch between them, the visual difference should be obvious.

Does that change work for you instead? Can we cherry-pick that commit into this branch?

mkernohanbc:
> The new focus style looks great!
> 
> One existing issue with this component is that our current markup and styles will cause some weirdness for the "Selection" story, as shown in your screenshot. The existing markup and style for the non-selected tag is causing its text and icon to sit visually higher in the container than it should be.
> 
> Here's a zoomed in example of what's in this branch currently: Selected TagGroup using this branch&apos;s code
> 
> I've added a [branch here](https://github.com/bcgov/design-system/tree/feature/refactor-tag-markup) where I've removed an unnecessary `<div>` from the markup and simplified the styling. This is what that version of the code looks like: Selection TagGroup with updated markup
> 
> If you open those two screenshots in separate tabs and switch between them, the visual difference should be obvious.
> 
> Does that change work for you instead? Can we cherry-pick that commit into this branch?

Looks good to me! Merged in [18cf561](https://github.com/bcgov/design-system/pull/396/commits/18cf5619a182dcc305872d092c88163079728eff)

ty2k (approved):

PR #400, opened 2024-06-28
Add TextArea component

mkernohanbc:
This PR expands on #389 to add a new Text Area component. 

Text Area is an alternative implementation of Text Field, designed for multi-line plain text input.

Changes from Text Field:

- replaces the <input> in the Text Field component with a <TextArea> ([as recommended by RAC](https://react-spectrum.adobe.com/react-aria/TextField.html#multi-line))
- removes the size and type variables
- removes the iconLeft and iconRight slots

Leaving this in draft for now, pending input from Philip on whether there should be any other design or behavioural variances between this and the base Text Field component.


mkernohanbc:
@ty2k documenting our conversation for reference: 

- We have[a case in the design spec](https://www.figma.com/design/2OhHsH2ztSFUwDcVzxug7Z/BC-Design-System?node-id=4192-4898&t=PopPihRWXO0CFRYu-4) for this component which displays a character counter.
- I've created a secondary `<span>` that will render and display the value of `maxLength` when that prop is present: 

Screenshot 2024-07-03 at 15 34 33

- I'm not sure what the best method is to capture and display the character count of the input `value`
- There doesn't seem to be a boilerplate solution to this within the RAC TextField or [useTextField](https://react-spectrum.adobe.com/react-aria/useTextField.html) APIs

mkernohanbc:
Flagging that the change in [c9c44b5](https://github.com/bcgov/design-system/pull/400/commits/c9c44b55efa88c0c93ac50e522408e724e989508) is throwing a TS error in the console, but doesn't actually seem to affect how the component renders in Storybook:

> `Type 'string | ((validation: ValidationResult) => string) | undefined' is not assignable to type 'ReactNode'.`
> `Type '(validation: ValidationResult) => string' is not assignable to type 'ReactNode'.`

mkernohanbc:
> Flagging that the change in [c9c44b5](https://github.com/bcgov/design-system/pull/400/commits/c9c44b55efa88c0c93ac50e522408e724e989508) is throwing a TS error in the console, but doesn't actually seem to affect how the component renders in Storybook:
> 
> > `Type 'string | ((validation: ValidationResult) => string) | undefined' is not assignable to type 'ReactNode'.`
> > `Type '(validation: ValidationResult) => string' is not assignable to type 'ReactNode'.`
> 
> Needs further investigation.

This is fixed in [a88ddc3](https://github.com/bcgov/design-system/pull/400/commits/a88ddc37d466360d83adab28506d32f6a8933836). The Typescript error seemed to originate with the way we were implementing the warning icon within the `FieldError`. To resolve this:

- removed `iconError` const from TextArea.tsx
- moved warning icon SVG into /src/assets/icon-error.svg
- use `::before` pseudo-class in TextArea.css to display the icon alongside the error message

mkernohanbc:
Just doing some cleanup to narrow the scope of this PR:

- [b6c025f](https://github.com/bcgov/design-system/pull/400/commits/b6c025f26a6781791f77c35df492afebc29a246f) and [8ef5d58](https://github.com/bcgov/design-system/pull/400/commits/8ef5d58f036b97c1bdbdf40e44850c46d630c2c8): removes WIP TextField code (see #389) used to prototype this component
- [b763129](https://github.com/bcgov/design-system/pull/400/commits/b7631293dc22529ce9807f54d56f54a0182d936e): removes test form from Vite app

mkernohanbc:
The character counter displayed when a `maxLength` is applied is now functional, thanks to @ty2k ([0b8bc37](https://github.com/bcgov/design-system/pull/400/commits/0b8bc37b0f6f5d972e9a64ba5683d6e474657786)). 

There is a small known issue with the character counter when a `defaultValue` prop is passed: in an uncontrolled input, the counter does not display the length of the `defaultValue` on render, but will update as soon as the input is modified.

This PR is ready for review.

mkernohanbc:
[7ae8882](https://github.com/bcgov/design-system/pull/400/commits/7ae8882cec3fa6ca88a2ecfe6b5c4c1496a8ee83) and [4e165df](https://github.com/bcgov/design-system/pull/400/commits/4e165df0c79234f1c9911244393716e9635278e8) contain some minor styling tweaks and expansion of the Storybook docs page.

packages/react-components/src/components/TextArea/TextArea.css
```diff
@@ -0,0 +1,130 @@
...
+
+/* Styles for the error message slot */
+.bcds-react-aria-TextArea--Error {
+  font: var(--typography-regular-small-body);
+  color: var(--typography-color-danger);
+}
+
+/* Warning icon */
+.bcds-react-aria-TextArea--Error::before {
+  content: url("../src/assets/icon-error.svg");
```
ty2k, line 126: Unfortunately this won't work as written when we ship this component because that `url()` SVG file referenced here won't exist in the code that's outputted by our `rollup` script. That line makes it way into the code, but there will be no such target SVG file at run time, so this will silently fail (because it's in CSS and not a JS import that's failing):

Error icon before reference in dist code

To solve, this is what I would do:

1. Remove `icon-error.svg` and instead add `src/components/SvgExclamation` similar to how we're shipping the `SvgBcLogo` now, as a React component that returns an `<svg>` tag.
    - Avoid the name `error` for this icon because it's potentially useful as a generic exclamation icon until it has other "error context" (red color, error message next to it, etc).
    - We want to avoid hard-coding `id` values in the SVG JSX if possible - see `SvgBcLogo` as an example. Most likely we only need an `id`on the `<svg>` parent tag, as a target for future automated tests.
    - To let people style the icon after the fact, we want to allow coloring the icon with CSS outside of the context of the SVG JSX file, so we should set `fill` to `currentColor`.
    - If you want to keep `icon-error.svg` in the assets folder as a reference to what `SvgExclamation` came from, we can do that like we've done with the other SVGs in that directory, but we should still make the `id` and `currentColor` changes in that copy of the file.
2. In the `TextArea` component file, pass the render prop `isInvalid` and use that to conditionally render a fragment `<></>` (or `<div>`, or whatever) around the `<SvgExclamation>` and the `<FieldError>{errorMessage}</FieldError>`. 
   - This should solve the issue you were having where `<FieldError>` can't have other `children`:
     FieldError TypeScript error

packages/react-components/src/stories/TextArea.stories.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+import { TextArea } from "../components";
+
+const meta = {
+  title: "Components/TextArea",
+  component: TextArea,
+  parameters: {
+    layout: "centered",
+  },
+  argTypes: {
```
ty2k, line 11: Let's add `errorMessage` in here like this:

```js
errorMessage: {
  control: { type: "text" },
  description: "Error message text that appears when `isInvalid` is true", // Or whatever you want
},
```

That will still give us the `union` type tag on the Storybook page, but we can have access to that field in the Controls block:

TextArea Storybook Controls block showing errorMessage field

packages/react-components/src/components/TextArea/TextArea.tsx
```diff
@@ -0,0 +1,85 @@
...
+              value={value}
+              onChange={(e) => handleChange(e.target.value)}
+            />
+          </div>
+          {description || maxLength ? (
+            <div className="bcds-react-aria-TextArea--Description">
+              {description && (
+                <Text
+                  slot="description"
+                  className="bcds-react-aria-TextArea--Description text"
```
ty2k, line 66: This class name and the class name for the counter are hard to understand because we're using `bcds-react-aria-TextArea--Description` for the parent. Let's be very explicit and rename this to just `text` and write the selector like `bcds-react-aria-TextArea--Description > text`, or use `bcds-react-aria-TextArea--Description-Text` or something.

ty2k (changes requested):
This rocks @mkernohanbc . Let me know if you want to pair on the SVG icon issue. No need to replace the other instances of that icon in this PR (in `Select` and `TextField`), but let's add the common version here to start.

packages/react-components/src/components/TextArea/TextArea.css
mkernohanbc: So, (1) is now done (see [1423f3e](https://github.com/bcgov/design-system/pull/400/commits/1423f3e45ce2f59b8cab83461543acc4d739e493), [95c8ac3](https://github.com/bcgov/design-system/pull/400/commits/95c8ac3d7cfed18eb8e58cf5970051d16dec9434) and [05f9a25](https://github.com/bcgov/design-system/pull/400/commits/05f9a253f237077bafb53b99cd6460a049f2c273)), and I made a quick attempt at (2), but seemed to immediately hit up against the same issue I was having before.

There seems to be an related/underlying issue here that if the `<FieldError>` is anything other than the direct child of its parent component (in this case, the `<TextField>`), it just stops working entirely. So when it's wrapped in a fragment or a div, the error message won't render at all. I encountered this when I was first trying to debug this, and it's what led me towards the SVG/CSS approach.

[RAC documentation](https://react-spectrum.adobe.com/react-aria/TextField.html#validation) is a little oblique on this, only says:

> To display validation errors, add a <FieldError> element as a child of the TextField.

packages/react-components/src/stories/TextArea.stories.tsx
mkernohanbc: Done in [40bfbdb](https://github.com/bcgov/design-system/pull/400/commits/40bfbdbcdfde0b1879b3a3373e9f147c8f142595)

packages/react-components/src/components/TextArea/TextArea.tsx
mkernohanbc: Done in [49038e3](https://github.com/bcgov/design-system/pull/400/commits/49038e3898f87f9ed629a9bb1f67fcd48966eb3f)

packages/react-components/src/components/TextArea/TextArea.css
mkernohanbc: Implemented in [d072d78](https://github.com/bcgov/design-system/pull/400/commits/d072d78dd24718c3ad214a2f7d91c13c4398d1fb)

mkernohanbc:
@ty2k this should now be good to go! 

Flagging that Storybook build issues I was experiencing this morning appeared to be fixed by bumping Storybook to 8.2.5 (see #411), but I've reverted those changes within this PR.

packages/react-components/src/components/SvgExclamationIcon/SvgExclamationIcon.tsx
```diff
@@ -0,0 +1,20 @@
...
+export default function SvgExclamationIcon({ id = "Exclamation icon" }) {
+  return (
+    <svg
+      id={id}
+      width="20"
+      height="20"
+      viewBox="0 0 20 20"
+      xmlns="http://www.w3.org/2000/svg"
+    >
+      <g id="20px/Exclamation icon">
```
ty2k, line 11: Can you please remove this `id` attribute and the one on `<path>` below, please? This is essentially just cruft from Figma or whatever app generated the SVG code and isn't used for anything, but attributes like this can conflict with genuine `id` attributes in the rest of the DOM. It should be removed from the SVG too.

ty2k (changes requested):
Awesome! We just need to clean up the SVG `id` stuff and this can be merged. :+1: :+1:

packages/react-components/src/components/SvgExclamationIcon/SvgExclamationIcon.tsx
mkernohanbc: Fixed in [3bbe419](https://github.com/bcgov/design-system/pull/400/commits/3bbe4191673f67eb26244096c4e8a77d27f12915)!

packages/react-components/src/components/SvgExclamationIcon/SvgExclamationIcon.tsx
```diff
@@ -0,0 +1,19 @@
+// This icon is available as a plain SVG at src/assets/icon-exclamation.svg
+export default function SvgExclamationIcon({ id = "Exclamation icon" }) {
```
ty2k, line 2: Sorry, just caught this - it's not valid to use whitespace in an HTML attribute (even if it works in most browsers), so this `id` prop default should become `exclamation-icon`.

ty2k (changes requested):

packages/react-components/src/components/SvgExclamationIcon/SvgExclamationIcon.tsx
mkernohanbc: Fixed in [cbe0370](https://github.com/bcgov/design-system/pull/400/commits/cbe0370c75e09322f59be838a08390d84fdc9cc9)

ty2k (approved):
Looks great!

PR #402, opened 2024-07-05
Storybook Nginx configuration for production deployment

ty2k:
This updates the React component library's Storybook app Nginx configuration to handle routing according to these rules:

- Requests to `/` will be redirected to the [B.C. Design System landing page on gov.bc.ca](https://gov.bc.ca/designsystem)
  - Ex: A GET request to `https://<design-system-subdomain>.gov.bc.ca` will go to BCDS page
- Requests to any valid address starting with `/react-components/` will get served the Storybook app, which handles its own routing client-side after the first page load
  - Ex: A GET request to `https://<design-system-subdomain>.gov.bc.ca/react-components/` will go to Storybook
  - Ex: `https://<design-system-subdomain>.gov.bc.ca/react-components/?path=/docs/components-button-button--docs` goes to Storybook's Button documentation page
- Requests to `/react-components` will be redirected to `/react-components/` (trailing slash)
- Requests that would result in 404 errors are redirected to the BCDS page

Keeping this Nginx configuration where it is makes sense for now, but it might need a re-think if we start serving multiple apps from one subdomain at different paths. In that case, it might make more sense to move the Nginx configuration to somewhere else in this repo, outside of `packages/react-components`.

### DIFF
--- a/packages/react-components/Dockerfile.storybook
+++ b/packages/react-components/Dockerfile.storybook
@@ -3,7 +3,7 @@
 # -----------
 # Build stage
 # -----------
-FROM node:lts-alpine as build-stage
+FROM node:lts-alpine AS build-stage
 ARG GITHUB_SHA
 ENV GITHUB_SHA=$GITHUB_SHA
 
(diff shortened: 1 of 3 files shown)


mkernohanbc (approved):
All makes sense to me, no notes!

ty2k:
We have provisioned `designsystem.gov.bc.ca` as the subdomain that Storybook will be served from.

I have created a Route object in OpenShift pointing that domain to our production deployment's Service object.

The Storybook site is currently being served from `https://designsystem.gov.bc.ca/`, and after this PR gets merged, that will change to `https://designsystem.gov.bc.ca/react-components/`.

PR #404, opened 2024-07-11
Fix Select button overflow issue

### LINKED ISSUE #403: Select - selecting a very long list option causes element to overflow
When using the Select component, if a very long option is selected, the Select `<button>` will expand to display the `SelectValue` on one line, even if it causes it to overflow its parent container. This is illustrated here:

Image

This issue only manifests when the over-long option is actually selected. In the list view, a long option will wrap properly:

Screenshot 2024-07-10 at 16 00 09

Expected behaviour would be either:

- The `SelectValue` wraps vertically
- The `SelectValue` is truncated

It looks like we have CSS that should handle this exact case in Select.css:

```
.bcds-react-aria-Select--Button > .bcds-react-aria-SelectValue {
  font: var(--typography-regular-body);
  overflow: hidden;
  white-space: nowrap;
  text-overflow: ellipsis;
  min-width: 0;
} 
```

I'm going to investigate and see if I can figure out why the `overflow` and `text-overflow` rules aren't being applied. I suspect that `white-space: nowrap` may be part of the issue here.

mkernohanbc:
This PR closes #403. It adds a `max-width` constraint to `bcds-react-aria-Select--Button`, to prevent the selector overflowing its container if a very long list item is selected. Now, if an over-length list option is selected, the truncation handling in Select.css applies correctly.

Before:
Screenshot 2024-07-11 at 10 38 34

After:
Screenshot 2024-07-11 at 10 42 45

This was tested in a 3-column grid layout using `react-grid-system` in our Vite app.

### DIFF
--- a/packages/react-components/package-lock.json
+++ b/packages/react-components/package-lock.json
@@ -10,8 +10,7 @@
       "license": "Apache-2.0",
       "dependencies": {
         "@bcgov/design-tokens": "3.0.0",
-        "react-aria-components": "1.2.1",
-        "react-grid-system": "^8.2.0"
+        "react-aria-components": "1.2.1"
       },
       "devDependencies": {
         "@react-stately/data": "3.11.4",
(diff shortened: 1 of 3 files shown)


ty2k (approved):
Great fix!

Before:
Select component with long option text overflowing the screen

After:
Select component with long option truncated

PR #401, opened 2024-07-03
Add Form component

mkernohanbc:
Moving this POC from #398 into a draft PR for discussion.

This PR adds a new Form component, based on [RAC Form](https://react-spectrum.adobe.com/react-aria/Form.html). An example instance and documentation page in Storybook is also included.

The Form component is intended to be used as a wrapper around other components, to provide additional support for data validation and submission and form styling. It renders an HTML `<form>` element. 

By default, the Form component renders without any styling of its own. The `className` prop can be used to style the form as a whole.

Some boilerplate CSS is included (see Form.css). This is currently only used to style the example form in Storybook.

PR #409, opened 2024-07-18
Add Form component

mkernohanbc:
Cherry-picked from #401 so that we can test with new components. Scope of this PR is the same:

This PR adds a new Form component, based on [RAC Form](https://react-spectrum.adobe.com/react-aria/Form.html). An example instance and documentation page in Storybook is also included.

The Form component is intended to be used as a wrapper around other components, to provide additional support for data validation and submission and form styling. It renders an HTML `<form>` element.

By default, the Form component renders without any styling of its own. The `className` prop can be used to style the form as a whole.

Some boilerplate CSS is included (see Form.css). This is currently only used to style the example form in Storybook.


mkernohanbc:
[c19a71d](https://github.com/bcgov/design-system/pull/409/commits/c19a71d3cd9594b327665d60e2cc06af70948100) simplifies the component structure:

- removes explicit prop declarations and CSS import from the component itself
- adds props as `argTypes` in Form.stories.tsx instead
- moves boilerplate CSS used for Storybook example into /src/stories
- expands documentation page in Storybook, in a separate 'Utility' section

This component is a convenience, that simply enables the user to import the RAC Form component without needing to add an additional dependency. 

With that, I think this is ready for review!

packages/react-components/src/stories/Form.mdx
```diff
@@ -0,0 +1,97 @@
...
+In this example, the Form component wraps around Text Field, Select and Button components to compose a simple form, which is styled using a boilerplate `.bcds-react-aria-Form` class.
+
+Button functionality is derived from the `type` prop.
+
+<Canvas of={FormStories.ComposedFormExample} />
+<Controls of={FormStories.ComposedFormExample} />
+
+### CSS for this example form
+
+<Source
```
ty2k, line 70: Instead of using the `<Source>` block, let's use a code fence so that the whitespace within the CSS snippet won't get mangled on save:

Screenshot 2024-07-23 at 1 41 54 PM

ty2k (changes requested):
Great addition! 🚀 Just one small change to the MDX file to make the code sample more readable.

When testing this component, I find the HTML validation is super easy to use. Features like passing `type="email"` to an input field or `type="reset"` to a button just work.

Example:

```tsx
<Form>
  <TextField label="Name" isRequired />
  <TextField type="email" label="Email" isRequired />
  <Select
    label="Employment status"
    items={[
      { id: "employee", label: "Employee" },
      { id: "contractor", label: "Contractor" },
    ]}
    isRequired
  />
  <Button type="submit" variant="primary">
    Submit
  </Button>
  <Button type="reset" variant="secondary">
    Reset
  </Button>
</Form>
```

After interacting with each field:
Validation error example

With valid data in each field:
Valid data example

Invalid email:
Invalid email example

packages/react-components/src/stories/Form.mdx
mkernohanbc: Fixed in [78cf1c9](https://github.com/bcgov/design-system/pull/409/commits/78cf1c9e3e4ed96d844fe1df584d83f048cd5c1d)!

ty2k (approved):
Looks great! ✅ 

Form component CSS

PR #412, opened 2024-07-22
Text Field - remove unnecessary padding rule from label

mkernohanbc:
Does what it says on the tin — nixes 2px worth of unnecessary padding from the label above a text field.


ty2k (approved):
This brings the TextField in line with the new TextArea #400.

PR #414, opened 2024-07-23
Select - visible label when required rather than optional

mkernohanbc:
This change changes how the `isRequired` prop is handled, to align with behaviour of TextField and TextArea. The goal is internal consistency and strict compliance with [WCAG SC 3.3.2](https://www.w3.org/WAI/WCAG21/Understanding/labels-or-instructions.html).

With this change, when `isRequired` is true, a secondary "(required)" label is rendered. When `isRequired` is false, only the primary label is displayed.

Old behaviour:
Screenshot 2024-07-23 at 10 25 47

New behaviour:
Screenshot 2024-07-23 at 10 25 26

### DIFF
--- a/packages/react-components/src/components/Select/Select.tsx
+++ b/packages/react-components/src/components/Select/Select.tsx
@@ -124,7 +124,7 @@ export default function Select<T extends object>({
         <>
           {label && (
             <Label className="bcds-react-aria-Select--Label">
-              {isRequired ? label : `${label} (optional)`}
+              {isRequired ? `${label} (required)` : label}
             </Label>
           )}
           <Button
--- a/packages/react-components/src/stories/Select.mdx
+++ b/packages/react-components/src/stories/Select.mdx
@@ -71,7 +71,7 @@ Choose from medium (default) and small select sizes:
 
 ### Required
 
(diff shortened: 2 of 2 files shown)


ty2k (approved):
Looks good. ✅ 

One thing we might want to consider for the future: Since we are now marking required instead of marking optional, we might want to add a red asterisk to the required state, if only as a backup option. It's super conventional, and on mobile the form labels are going to get long with all of them marked with "(required)" now. If folks try to implement an asterisk themselves with our components, they won't be able to do it via props (at least not without the TypeScript compiler complaining). They'll have to set the asterisk outside of the component and line it up with CSS.

mkernohanbc:
> Looks good. ✅
> 
> One thing we might want to consider for the future: Since we are now marking required instead of marking optional, we might want to add a red asterisk to the required state, if only as a backup option. It's super conventional, and on mobile the form labels are going to get long with all of them marked with "(required)" now. If folks try to implement an asterisk themselves with our components, they won't be able to do it via props (at least not without the TypeScript compiler complaining). They'll have to set the asterisk outside of the component and line it up with CSS.

I'm definitely open to revisiting how we handle this behaviour post-release, as long as we're consistent across components!

PR #419, opened 2024-07-24
0.1.0 release docs

mkernohanbc:
This PR aggregates content updates to various docs pages in Storybook and on GitHub to go live alongside the component library v0.1.0 release.

### DIFF
--- a/packages/react-components/src/stories/Introduction.mdx
+++ b/packages/react-components/src/stories/Introduction.mdx
@@ -1,4 +1,4 @@
-import { Meta, Source } from "@storybook/blocks";
+import { Meta, Source, Subtitle } from "@storybook/blocks";
 
 import packageJson from "../../package.json";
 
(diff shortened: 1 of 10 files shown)


packages/react-components/src/stories/Introduction.mdx
```diff
@@ -15,9 +15,14 @@ export const link =
 
 <Meta title="Introduction" />
 
-# B.C. Design System React Components
+# B.C. Design System React Components v{packageJson.version}
```
ty2k, line 18: This new chunk **should** always work but let's write it like this instead to make sure:

```js
{packageJson?.version ? `v${packageJson.version}` : null}
```

packages/react-components/src/stories/Introduction.mdx
```diff
@@ -15,9 +15,14 @@ export const link =
...
 
-This is a UI workshop for React components included in the [B.C. Design System](https://gov.bc.ca/designsystem). This app provides:
+<Subtitle>
```
ty2k, line 20: Looks good!

Storybook introduction page

ty2k (changes requested):
Looks awesome! One small change request for the way we inject the package.json version.

packages/react-components/src/stories/Introduction.mdx
mkernohanbc: Fixed in [993b7e0](https://github.com/bcgov/design-system/pull/419/commits/993b7e0e7094d5ee9b3953408d80c4db85c5c09f)!

ty2k (approved):
🚀

PR #429, opened 2024-07-25
TagGroup - add warning icon to error message

mkernohanbc:
This change adds a visible warning icon alongside the `errorMessage` displayed when TagGroup is in an invalid state. This mirrors behaviour from the TextArea component, and is intended to improve conformance with [WCAG SC 1.4.1](https://www.w3.org/WAI/WCAG21/Understanding/use-of-color.html)

Before:
Screenshot 2024-07-25 at 09 08 59

After:
Screenshot 2024-07-25 at 09 08 49

### DIFF
--- a/packages/react-components/src/components/TagGroup/TagGroup.css
+++ b/packages/react-components/src/components/TagGroup/TagGroup.css
@@ -11,5 +11,13 @@
 }
 
 .bcds-react-aria-TagGroup--Text-error {
+  display: flex;
   color: var(--typography-color-danger);
+}
+
+.bcds-react-aria-TagGroup--Text-error > svg {
+  padding-right: var(--layout-padding-xsmall);
+  color: var(--icons-color-danger);
+  width: var(--icons-size-medium);
+  height: var(--icons-size-medium);
 }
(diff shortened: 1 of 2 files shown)


mkernohanbc:
@Philip-Cheung also tagging you for review — if you approve of this change, let's mirror it on the Figma component, please!

This change won't ship immediately, we'll bundle it into our next release.

packages/react-components/src/components/TagGroup/TagGroup.css
```diff
@@ -11,5 +11,13 @@
...
 
 .bcds-react-aria-TagGroup--Text-error {
+  display: flex;
   color: var(--typography-color-danger);
 }
+
+.bcds-react-aria-TagGroup--Text-error > svg {
+  padding-right: var(--layout-padding-xsmall);
+  color: var(--icons-color-danger);
+  width: var(--icons-size-medium);
```
ty2k, line 21: Inside of a flex container, `width` has potential to collapse depending on the other children of the container. Here's what happens with a long error message:

TagGroup with icon using width

If you use `min-width` instead and keep the `height` as it is, you should get what you're looking for:

TagGroup with icon using min-width

ty2k (changes requested):
Looks great, one minor change to the CSS for the icon needed.

packages/react-components/src/components/TagGroup/TagGroup.css
mkernohanbc: Fixed in [883c458](https://github.com/bcgov/design-system/pull/429/commits/883c458cc2a17a04845b882a7ea3a0eb436d6e6b), flagging that we might need to refactor this on a couple of other components too!

ty2k (approved):

PR #431, opened 2024-07-25
Select - expose description field

mkernohanbc:
This PR adds an optional description/helper text field to the Select component, rendering below the selector input. The styling and behaviour of this description field follows the pattern of our other input components:

Screenshot 2024-07-25 at 10 36 33

@Philip-Cheung as with #429, if you're happy with this let's also add this to the Figma component in our next release, please!

### DIFF
--- a/packages/react-components/src/components/Select/Select.css
+++ b/packages/react-components/src/components/Select/Select.css
@@ -17,6 +17,13 @@
   color: var(--typography-color-disabled);
 }
 
+/* Rext description below select input */
+.bcds-react-aria-Select--Description {
+  font: var(--typography-regular-small-body);
+  color: var(--typography-color-secondary);
+  padding: var(--layout-padding-xsmall) var(--layout-padding-none);
+}
+
 /* Error message */
 .bcds-react-aria-Select--Error {
   font: var(--typography-regular-small-body);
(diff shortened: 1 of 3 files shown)


ty2k (changes requested):
Great addition. While we're updating this, can we get the `errorMessage` explicitly added to the `argTypes` in the stories file so that it's easy to drop a string in for testing?

```js
errorMessage: {
  control: { type: "text" },
  description: "Error message text that appears when `isInvalid` is true",
},
```

Otherwise all good to merge IMO:
Storybook screenshot

mkernohanbc:
> Great addition. While we're updating this, can we get the `errorMessage` explicitly added to the `argTypes` in the stories file so that it's easy to drop a string in for testing?
> 
> ```js
> errorMessage: {
>   control: { type: "text" },
>   description: "Error message text that appears when `isInvalid` is true",
> },
> ```
> 
> Otherwise all good to merge IMO: Storybook screenshot

Added in [fe9e870](https://github.com/bcgov/design-system/pull/431/commits/fe9e870bc45202f326dd6eeb2b8c0b17992f761b)!

ty2k (approved):
LGTM! 🚀

PR #437, opened 2024-08-01
Add inline alert component

mkernohanbc:
This change adds a new Inline Alert component, along with Storybook docs and examples on the Vite kitchen sink app:

Screenshot 2024-08-01 at 13 42 01

An inline alert is used to alert the user to important information or messages. It renders a styled `<div>` with [the ARIA "note" role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/note_role). ~~Optionally, the user can pass the `alertRole` prop to instead assign [the "alert" role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/alert_role).~~ 

**Update**: `alertRole` has been replaced by generic `role` prop in [4def3fa](https://github.com/bcgov/design-system/pull/437/commits/4def3fabaf396923760fd77d3bde240640cf9994), [see explanation here](https://github.com/bcgov/design-system/pull/437#issuecomment-2274380033)

The inline alert component has 4 visual themes, set using the `variant` prop. Applying a theme changes the alert colour scheme and the icon displayed on the left:

- `info` (default, blue theme)
- `success` (green)
- `warning` (orange)
- `error` (red)

The `title` and `description` props are used to populate the alert content. `hideIcon` toggles off the alert icon on the left side. `isCloseable` is used to render a 'close' icon button on the right side.

The component also includes a generic `children` slot, into which users can pass other components. The primary intended usage for `children` is to add Button components for additional actions the user can take.

This PR also implements 4 new icon components (`SvgCheckCircleIcon`, `SvgExclamationCircleIcon`, `SvgCloseIcon`, and `SvgInfoIcon`), used by the different alert themes.


mkernohanbc:
Leaving this PR in draft pending 2 outstanding pieces of work:

- [ ] Figure out what the `onPress` behaviour for the close button should be, and documenting its configuration
- [ ] Fixing erratic positioning of the theme icons, so that they are properly aligned with the alert title

@ty2k can we pair on these issues when you're next in DS mode, please?

mkernohanbc:
I've made a substantive change to how ARIA roles are configured in [4def3fa](https://github.com/bcgov/design-system/pull/437/commits/4def3fabaf396923760fd77d3bde240640cf9994).

Alert will still render with the "note" role by default, but now the `alertRole` prop is replaced with a generic `role` prop that can accept any valid ARIA role. This should be more flexible.

Docs have been updated to reflect this change, and recommend that users pick from these 4 roles:

- note (default)
- complementary
- status
- alert

mkernohanbc:
[8d7979e](https://github.com/bcgov/design-system/pull/437/commits/8d7979e6aaafb9e84d6941199e0879d40350db2e) is a hygiene change that touches a bunch of different files across the repo.

Purpose of this commit is to declutter the /src/components directory by moving all of the SVG icon components into a separate subfolder. This also necessitates updating imports in other components that use these icons.

packages/react-components/src/components/Select/Select.css
```diff
@@ -17,6 +17,13 @@
   color: var(--typography-color-disabled);
 }
 
+/* Rext description below select input */
```
ty2k, line 20: Typo `Rext`

ty2k (changes requested):
This is awesome. Some minor changes are needed to existing stuff and I have some thoughts on possible changes:

## Name

No blocker to merging, just my thoughts on the name:
- Why "Inline" Alert?
- I see React Spectrum has [InlineAlert](https://react-spectrum.adobe.com/react-spectrum/InlineAlert.html) but no equivalent RAC component. If they add an "InlineAlert", do we break our existing API to follow theirs?
- Other systems:
  - MUI has [Alert](https://mui.com/material-ui/react-alert/)
  - Ant Design has [Alert](https://ant.design/components/alert)
  - Bootstrap has [Alert](https://getbootstrap.com/docs/4.0/components/alerts/)
  - Chakra has [Alert](https://v2.chakra-ui.com/docs/components/alert)
  - Tailwind doesn't have pre-defined alert components, but their docs refer to the visual example as [Alerts](https://v1.tailwindcss.com/components/alerts)
  - Core UI has [Alert](https://coreui.io/react/docs/components/alert/)
- What if we called this just "Alert" for now?
  - If RAC releases "InlineAlert", we can adopt it with their same API rather than having to break our own component's API. If it just happens that they use the same API we are choosing with this component, we can easily add an alias for "InlineAlert".
  - If "Alert" is too generic, what about "AlertInline" to help the grouping with "AlertBanner"?

## Props

### `children`

I like `title` and `description` as props for being able to quickly do the most common "bold title and one line of text" look. However, I feel like `children` needs to be a generic slot that replaces both if it's present. Right now, `children` is good for holding a group of buttons, but we need to consider cases like:
  - Description text with inline elements in the text, like bold/italic tags and links.
This will make our API more like what React Spectrum is doing for `InlineAlert` which will give us a better chance of API compatibility if they do a generic version of `InlineAlert` in RAC.
  - Two lines of description text.
 
If the action button group is part of this design and present in other components, maybe we add a specific button group flex container component with that same spacing for use in compositions like this? Ex:

```jsx
<Alert>
  <ButtonGroup>
    <Button variant="secondary">Secondary</Button>
    <Button variant="primary">Primary</Button>
  </ButtonGroup>
</Alert>
```

### `hideIcon`

Generally speaking, I think we should always use `is` for boolean props. Or `has`/`should`, in rare cases. I think this is generally what RAC is doing. The exceptions they have to that rule are usually in cases where they are passing the name of an existing HTML attribute. For example, look at the [InlineAlert prop `autoFocus`](https://react-spectrum.adobe.com/react-spectrum/InlineAlert.html#props) which becomes the [HTML attribute `autofocus`](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/autofocus).

I would name this `isIconHidden`.

### Close handling

I think an `onClose` function prop should be added, to be accessed from the `<Button>` like `onPress={onClose}`. Users should be responsible for handling what `onClose` does outside of this component - we shouldn't attempt to handle any state inside of it ourselves. The most likely case I can see is some version of:

```jsx
function Main() {
  const [isAlertShown, setIsAlertShown] = useState(true);

  return (
    <main>
      {isAlertShown && (
        <Alert onClose={() => setIsAlertShown(false)}>
      )}
      <h1>Topic</h1>
      <p>Content</p>
    </main>
  )
}
```

packages/react-components/src/components/Select/Select.css
mkernohanbc: This is fixed in [9f3973c](https://github.com/bcgov/design-system/pull/437/commits/9f3973cf94aabf261f4181b43e577c374bd12556)

packages/react-components/src/components/InlineAlert/InlineAlert.css
```diff
@@ -0,0 +1,104 @@
...
+  color: var(--icons-color-primary);
+}
+
+.bcds-Inline-Alert--closeIcon > svg {
+  min-width: var(--icons-size-medium);
+  height: var(--icons-size-medium);
+}
+
+/* Info variant */
+.info {
```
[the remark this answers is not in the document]
mkernohanbc: This is fixed in [9f3973c](https://github.com/bcgov/design-system/pull/437/commits/9f3973cf94aabf261f4181b43e577c374bd12556)

mkernohanbc:
> ### `hideIcon`
> Generally speaking, I think we should always use `is` for boolean props. Or `has`/`should`, in rare cases. I think this is generally what RAC is doing. The exceptions they have to that rule are usually in cases where they are passing the name of an existing HTML attribute. For example, look at the [InlineAlert prop `autoFocus`](https://react-spectrum.adobe.com/react-spectrum/InlineAlert.html#props) which becomes the [HTML attribute `autofocus`](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/autofocus).
> 
> I would name this `isIconHidden`.

This is changed to `isIconHidden` in [9f3973c](https://github.com/bcgov/design-system/pull/437/commits/9f3973cf94aabf261f4181b43e577c374bd12556)

mkernohanbc:
> ## Props
> ### `children`
> I like `title` and `description` as props for being able to quickly do the most common "bold title and one line of text" look. However, I feel like `children` needs to be a generic slot that replaces both if it's present. Right now, `children` is good for holding a group of buttons, but we need to consider cases like:
> 
> * Description text with inline elements in the text, like bold/italic tags and links.
>   This will make our API more like what React Spectrum is doing for `InlineAlert` which will give us a better chance of API compatibility if they do a generic version of `InlineAlert` in RAC.
> * Two lines of description text.
> 
> If the action button group is part of this design and present in other components, maybe we add a specific button group flex container component with that same spacing for use in compositions like this? Ex:
> 
> ```js
> <Alert>
>   <ButtonGroup>
>     <Button variant="secondary">Secondary</Button>
>     <Button variant="primary">Primary</Button>
>   </ButtonGroup>
> </Alert>
> ```

[5e93023](https://github.com/bcgov/design-system/pull/437/commits/5e9302378131d4a86bd73efc4eec369ec00ae4d3) includes:

- Reworking of `children` slot to override `title` and `description`
- Removal of additional actions examples from Storybook
- Updates to Storybook docs to explain usage of `onClose` and `children`

I've created a new issue (#443) to work up the ButtonGroup idea. Once we have that, I'll revisit our example stories.

ty2k (approved):
Great stuff, merge away!

Issue #443, opened 2024-08-09
ButtonGroup


mkernohanbc:
New component idea (see @ty2k comment in #437 for context):

ButtonGroup is a utility component similar to Form, intended to help users easily create more complex UI compositions.

Its purpose is to provide a simple styled container into which the user can pass multiple button components, to be rendered horizontally alongside one another.

ButtonGroup will be useful as a subcomponent for:

- Header
- Form
- Inline Alert
- Future content components like Card and Callout

PR #446, opened 2024-08-12
Add ButtonGroup component and refactor Inline Alert

mkernohanbc:
This change adds a new `ButtonGroup`, as described in #443. 

ButtonGroup renders a `<div>` with the [ARIA "group" role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/group_role). It expects button components to be passed as `children`.

ButtonGroup has props to configure its layout:

- `orientation` can be set to `horizontal` or `vertical`
- `alignment` can be set to `start`, `center` or `end`

The default layout is a horizontal row, with the buttons pinned to the left side (`orientation="horizontal"`, `alignment="start"`)

[47e8b38](https://github.com/bcgov/design-system/pull/446/commits/47e8b38b7e69d5858594219296ce05df6e864a1a) refactors the Inline Alert component to include a new `buttons` slot. The user can pass an array of button components via this prop, that will then be rendered inside a ButtonGroup. An example story is included in both the ButtonGroup and InlineAlert docs:

Screenshot 2024-08-12 at 16 48 00

This PR also includes some other minor changes to Inline Alert, primarily focused on the Storybook examples and docs and fixing a broken `aria-labelledby` property ([0c5b596](https://github.com/bcgov/design-system/pull/446/commits/0c5b5966bfde21d2b75eb34ff62f1285bfa25b19))

### DIFF
--- a/packages/react-components/src/App.tsx
+++ b/packages/react-components/src/App.tsx
@@ -8,6 +8,7 @@ import { Button, Footer, FooterLinks, Header } from "@/components";
 import useWindowDimensions from "@/hooks/useWindowDimensions";
 import {
   ButtonPage,
+  ButtonGroupPage,
   InlineAlertPage,
   SelectPage,
   TagGroupPage,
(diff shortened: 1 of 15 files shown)


ty2k (approved):
Wicked. Can't wait to never have to write this code again. Ship it. 🚀

Issue #444, opened 2024-08-09
Inline alert - fix close button positioning


mkernohanbc:
The initial implementation of this component in #437 has a minor issue where the close button positioning is off-kilter relative to the title and icon fields.

Going to take this issue into a new branch to fix.

PR #449, opened 2024-08-13
InlineAlert close button is removed from document flow to center alert text vertically

ty2k:
Marking this as a draft so we can discuss whether we prefer to go this route or to make a smaller version of `Button`. See #444.

This PR updates the `InlineAlert` to remove the close button from the document flow, positioning it with `position: absolute` and a negative `top` value to center it.

The advantage of this approach is that we retain the 32px x 32px click target size for the close button, rather than rolling a smaller button. The disadvantage is that the code is more hacky and less declarative than using a smaller button and keeping the current `InlineAlert` code.

Single line of text, before and after:
Single line, before
Single line, after

Multiple lines of text, before and after:
Multi line, before
Multi line, after

### DIFF
--- a/packages/react-components/src/components/InlineAlert/InlineAlert.css
+++ b/packages/react-components/src/components/InlineAlert/InlineAlert.css
@@ -35,11 +35,30 @@
   height: var(--icons-size-medium);
 }
 
-/* Close icon button */
+/* Closeable */
+.bcds-Inline-Alert.closeable {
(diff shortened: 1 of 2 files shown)


mkernohanbc (approved):

ty2k:
Marking this as ready for review after our team discussion. Let's go forward with this plan for now.

PR #450, opened 2024-08-13
Select - refactor icons

mkernohanbc:
This change updates the Select component to use componentised versions of icons, rather than instantiating icons as functions within the component itself.

- fac327c replaces `iconError` with the `SvgExclamationIcon` component, and adds some additional CSS to style it
- c4ecae4 adds two new icon components (`SvgChevronUpIcon` and `SvgChevronDownIcon`)
- ca25172 replaces `ChevronUp` and `ChevronDown` functions with the new components


packages/react-components/src/components/Select/Select.tsx
mkernohanbc, line 1: Do we need the React import in line 1, or is that legacy stuff?

mkernohanbc, line 1: @ty2k do we need the React import on line 1, or is that legacy stuff?

ty2k: We do need it in this case, because we're using it for our TypeScript definitions, like `iconLeft?: React.ReactElement;`.

ty2k:
Couple things with this one:

In our component files, we don't want be to using the [Vite `resolve.alias`](https://vitejs.dev/config/shared-options#resolve-alias) `@` path when we're importing other files. This causes circular dependency warnings when we run the Rollup script that builds the component library for publishing. In general, we don't want to introduce new warnings in this process's log to keep it useful for us. The existing warnings we have in this output are from our `node_modules` dependencies which we don't have direct control over. Below is an example where the `InlineAlert` is causing a warning. I will push a commit to this branch that fixes this in both the `InlineAlert` and `Select`.

Terminal output of Rollup script showing warning for circular dependency on InlineAlert

---
 
Secondly, we should avoid extraneous `<span>` elements around the SVGs where possible. In this case we are able to avoid it. I will push a commit to this branch that fixes this.

ty2k:
`Select` before and after (no extra spacing on the exclamation icon by avoiding the `<span>`)

Before
Select before this PR
Select after this PR

`InlineAlert` before and after (no changes):
InlineAlert before this PR
InlineAlert after this PR

ty2k (approved):

PR #454, opened 2024-08-19
Test design tokens build

ty2k:
This PR adds a test script that uses the [Node.js test runner](https://nodejs.org/docs/latest-v20.x/api/test.html) to run the build script from `build-output.js`. For now, this only tests that the script completes without errors - it doesn't make any assertions about the different contents of the build directory. The new test script is added to the design tokens `package.json` file, and a GitHub Actions workflow is added that runs the test script when a PR updates files within the `./packages/design-tokens` directory.

The goal of this is to allow us to more safely update the design tokens package and to give us a test framework that we can add new assertions to over time (correctness of select tokens, presence of the expected output formats, etc).

### DIFF
--- a/.github/workflows/test_design_tokens_build.yaml
+++ b/.github/workflows/test_design_tokens_build.yaml
@@ -1,0 +1,38 @@
+name: Test design tokens build script
+
+on:
+  pull_request:
+    paths:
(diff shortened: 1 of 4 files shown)


mkernohanbc (approved):
lgtm!

PR #469, opened 2024-08-23
Update components to use 3.1.0 tokens

mkernohanbc:
This PR does some cleanup on the tokens used in various components, fixing some inconsistent usages and aligning with new tokens in 3.1.0: 

- c816270 and c8e916f: Switch
- 9ba8d67 and c652c95: Select
- f0dedc2: Header
- 01f936a: Tag
- 959a1df and 614f81a: TextField
- bab8635: TextArea

### DIFF
--- a/packages/react-components/src/components/Select/Select.css
+++ b/packages/react-components/src/components/Select/Select.css
@@ -120,13 +120,13 @@
 
 /* ListBox option item */
 .bcds-react-aria-Select--ListBoxItem {
-  border-radius: var(--layout-margin-xsmall);
+  border-radius: var(--layout-border-radius-medium);
   cursor: pointer;
   display: flex;
   flex-direction: row;
   align-items: center;
   gap: var(--layout-margin-small);
-  padding: 6.5px var(--layout-padding-small);
+  padding: var(--layout-padding-small);
 }
 .bcds-react-aria-Select--ListBoxItem[data-focused],
 .bcds-react-aria-Select--ListBoxItem[data-hovered] {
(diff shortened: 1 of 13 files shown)


mkernohanbc:
@ty2k leaving this in draft for the moment, expecting there may be more commits to come on this.

mkernohanbc:
@ty2k I'm done with this branch now. Over to you for review, or feel free to use this branch to fold in any other revisions on your backlog!

packages/react-components/src/components/Select/Select.css
```diff
@@ -115,18 +115,18 @@
...
 /* ListBox option item */
 .bcds-react-aria-Select--ListBoxItem {
-  border-radius: var(--layout-margin-xsmall);
+  border-radius: var(--layout-border-radius-medium);
   cursor: pointer;
   display: flex;
   flex-direction: row;
   align-items: center;
   gap: var(--layout-margin-small);
-  padding: 6.5px var(--layout-padding-small);
```
ty2k, line 129: I see this matches the current version of the component in Figma, but just flagging that this will cause the list of items to be physically taller:

Before:
Screenshot 2024-08-27 at 10 05 26 AM

After:
Screenshot 2024-08-27 at 10 05 29 AM

ty2k (approved):
:+1: :+1:

packages/react-components/src/components/Select/Select.css
mkernohanbc: Noted! I'm good with this if you are? My view is prioritising consistency between Figma/React here.

ty2k: Yep, I agree with your priority there, no changes!

PR #476, opened 2024-08-27
Unit tests for Header, Footer, Form, InlineAlert, Switch

ty2k:
This adds unit tests for:

- Header f3b36c2e3a7e1ff721a670d6cd759e85464329e1
- Footer 849216f71b322b768567506709668ed97c496339
- Form 7066f14ac50aa55ad2f73db0c2defb2a989147a8
- InlineAlert 30d8e27780948de921e31876833fdea555265b9d
- Switch c43e32c652ffa992778d0a6cd8239c277e4b5217

I found that some existing default React import statements (`import React from "react"`) caused tests to not run. In cases where we were using methods on the `React` object, I have explicitly kept those imports. In cases where we were only using `React` for type information, I have removed the default import statement. This shouldn't make any difference - it's just to keep the TypeScript compiler happy. @mkernohanbc you had mentioned this in #450 and at that time, I couldn't remove the import statement without the compiler complaining on those `React.ReactElement` type fields. We haven't changed our `tsconfig.json` file, so I'm thinking something may have changed in one of the underlying devDependencies. 🤷‍♂️ 

Speaking of the TypeScript compiler, I find I can't add a test file for the Select component because of implicit types that the Jest test runner feels are missing:

Jest output for Select test

I'm not sure what the best way to fix that is yet, so there's no test file added for Select in this PR.

All tests should be passing in this PR.

### DIFF
--- a/packages/react-components/src/components/Footer/Footer.test.tsx
+++ b/packages/react-components/src/components/Footer/Footer.test.tsx
@@ -1,0 +1,16 @@
+import "@testing-library/jest-dom";
+import { render, screen } from "@testing-library/react";
+
+import Footer from "./Footer";
+
+describe("Footer", () => {
+  render(<Footer />);
+
+  // https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/contentinfo_role
+  const contentInfoElements = screen.getAllByRole("contentinfo");
+
+  it("renders an HTML footer component", () => {
+    expect(contentInfoElements).toHaveLength(1);
+    expect(contentInfoElements[0].tagName).toBe("FOOTER");
+  });
+});
--- a/packages/react-components/src/components/Footer/Footer.tsx
+++ b/packages/react-components/src/components/Footer/Footer.tsx
@@ -1,5 +1,3 @@
-import React from "react";
-
 import SvgBcLogo from "../Icons/SvgBcLogo";
 
 import "./Footer.css";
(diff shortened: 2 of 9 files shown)


mkernohanbc (approved):
nothing but thumbs over here

mkernohanbc:
> Speaking of the TypeScript compiler, I find I can't add a test file for the Select component because of implicit types that the Jest test runner feels are missing:
> 
> Jest output for Select test
> I'm not sure what the best way to fix that is yet, so there's no test file added for Select in this PR.
> 
> All tests should be passing in this PR.

This is weird, because all of those props have type declarations. If we explicitly re-declare/type those props within `SelectProps`, would it work (and is there a disadvantage to doing that beyond "this _should_ be redundant"?)

ty2k:
> > Speaking of the TypeScript compiler, I find I can't add a test file for the Select component because of implicit types that the Jest test runner feels are missing:
> > Jest output for Select test
> > I'm not sure what the best way to fix that is yet, so there's no test file added for Select in this PR.
> > All tests should be passing in this PR.
> 
> This is weird, because all of those props have type declarations. If we explicitly re-declare/type those props within `SelectProps`, would it work (and is there a disadvantage to doing that beyond "this _should_ be redundant"?)

Yeah, I can see the type information in VS Code when I hover those variables. It's the subtle differences in environment between VS Code's idea of the types at code time vs the TypeScript compiler's idea of the types at compile time vs Jest's idea of the types at test time. I don't understand the difference yet. Probably the solution will look like importing the types from the underlying RAC libraries (maybe `react-aria-components`, maybe one of the underlying state libraries like `@react-stately`) and then those render prop lines will look something like...

```tsx
{({ isOpen, isRequired, isInvalid }: SelectStateProps) => (
```

... or whatever. I can't get it working yet! :)

Issue #500, opened 2024-11-18
Update React Aria Components to 1.4.1


mkernohanbc:
The `design-system-react-components` package currently relies on `react-aria-components@1.3.3`. Since our last release, Adobe have released a new minor version of RAC (1.4.x), and are now at [v1.4.1](https://www.npmjs.com/package/react-aria-components).

The components added in [#499](https://github.com/bcgov/design-system/pull/499) rely on features introduced in RAC 1.4.0, so we will need to update RAC as part of our 0.5.0 release.

mkernohanbc:
Flagging that RAC has [moved to 1.5.0 as of last week](https://react-spectrum.adobe.com/releases/2024-11-20.html), and the new components used in #499 are now generally available.

PR #504, opened 2024-11-27
Refactor Select to use ListBoxSection

### LINKED ISSUE #503: Update Select to use ListboxSection instead of Section
The [latest version of React Aria Components](https://react-spectrum.adobe.com/releases/2024-11-20.html) deprecates the generic `Section` component, in favour of "specific components for each parent collection component, e.g. ListBoxSection and MenuSection".

This affects our Select component, which uses `Section`. We will need to refactor this component to properly support `ListBoxSection` alongside #500.

mkernohanbc:
This PR closes #503 with a small change to the Select component, replacing the RAC Section subcomponent (deprecated in 1.5.0) with the ListBoxSection subcomponent. This change is dependent on #500.

Section was only used internally within Select (ie. our users are not importing/working with Section directly), so there should be no downstream impact for users from this change. The component is configured and will render as before.

### DIFF
--- a/packages/react-components/package-lock.json
+++ b/packages/react-components/package-lock.json
@@ -10,7 +10,7 @@
       "license": "Apache-2.0",
       "dependencies": {
         "@bcgov/design-tokens": "3.1.1",
-        "react-aria-components": "1.3.3"
+        "react-aria-components": "1.5.0"
       },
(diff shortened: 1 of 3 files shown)


mkernohanbc:
Holding on merging until #500 is resolved.

PR #505, opened 2024-11-27
Tags: enhancement to incorporate chips styling

Supriya-Arora:
Added style (rectangular/circular) & size (small/medium) prop to Tag component.

image


packages/react-components/src/components/Tag/Tag.tsx
```diff
@@ -25,12 +25,27 @@ export interface TagProps extends ReactAriaTagProps {
     | "green"
     | "red"
     | "yellow";
+  /**
+   * Style
+   */
+  style?: "rectangular" | "circular";
```
mkernohanbc, line 31: The `style` prop name [already exists](https://react-spectrum.adobe.com/react-aria/TagGroup.html#tag) on RAC Tag, providing a method for injecting inline CSS into a component. 

I'd suggest we rename this to `variant` (following the pattern of our other components, like Button.)

mkernohanbc (changes requested):
I think this looks good structurally, and all seems to work as expected locally. The only issue I see is the prop name collision I flagged in my other comment. Once we update the `style` prop name and references to resolve that conflict, I'll approve and we can merge! 

One thing I'd like to circle back on as another enhancement is extending `TagGroup` so that the user can set the tag style once and have it be applied to all child tags in that group (ie. as an alternative to setting the style on each tag individually.) I'll create a separate ticket for that change — not blocking for release, just a QOL thing.

mkernohanbc (approved):

mkernohanbc:
@Supriya-Arora looks good to me! 🚀

PR #510, opened 2024-12-12
Add Heading and Text components

mkernohanbc:
This PR implements the RAC Heading and Text utility components into the B.C. Design System:

Screenshot 2024-12-12 at 10 49 01

These components are intended to be used internally within other components, and to provide a straightforward way for users to compose sections of content within their interfaces using the BCDS typescale.

Both components are styled. The `color` prop enables the user to set text colour, from the design system's colour palette. Content is populated via the `children` slot.

The user can override styling using these props:

- `className`: replaces BCDS CSS class
- `isUnstyled`: removes CSS classes entirely, allowing components to inherit their styling from parent elements

## Heading
Heading also accepts these props:

- `level`: used to set the heading type, from `h1` to `h6`

## Text
Text also accepts these props:

- `elementType`: sets the type of HTML element rendered (defaults to `span`)
- `size`: sets the text size (defaults to `medium`, also includes `large` and `small`)


mkernohanbc:
> For Text component, it will be great if we can handle error scenario when elementType is blank/null. 

Fixed in [5504f6f](https://github.com/bcgov/design-system/pull/510/commits/5504f6f973efb6ad759862761a60a0073ed5fb2e)!

PR #516, opened 2025-01-20
SCSS format support

plasticviking:
Add the StyleDictionary built in support for SCSS variables generation to the build script. Some frameworks (Bootstrap, and Quarto, at least) benefit from the flexibility of SCSS variables versus CSS.

### DIFF
--- a/packages/design-tokens/build-output.js
+++ b/packages/design-tokens/build-output.js
@@ -226,6 +226,30 @@ const sd = new StyleDictionary({
         },
       ],
     },
+    scss: {
+      transformGroup: "tokens-studio",
+      transforms: [
+        "bcds/typography/italic",
+        "bcds/color/uppercase",
+        "bcds/css/size/zero",
+        "name/kebab",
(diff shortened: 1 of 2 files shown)


packages/design-tokens/build-output.js
```diff
@@ -226,6 +226,30 @@ const sd = new StyleDictionary({
         },
       ],
     },
+    scss: {
```
mkernohanbc: Can we please add an additional `scssPrefixed` transform, following the pattern of our other formats?

dawnbcgov:
Hi @mkernohanbc -- @plasticviking was a vendor no longer working on the project that generated this PR.

I would say, if they don't respond within a day or two indicating they'd like to make the change I'd love for you to make the requested changes on your end and merge it! The timing is good because we may want to update our scss file after your next release and have some planned upcoming work :)

plasticviking:
I've made the requested changes, but do not have merge access

PR #551, opened 2025-11-19
Migrate npm publishing workflows to OIDC

mkernohanbc:
Draft PR for discussion.

This PR will update our workflows to use OIDC instead of access tokens to authenticate when publishing to npm:

- 693ba83 and 937921a add [the necessary permissions](https://docs.npmjs.com/trusted-publishers#github-actions-configuration) to our two publishing workflows
- 95f740a removes existing usages of npm access tokens from the workflows

The outstanding piece is how we redesign these workflows to:

1. Continue publishing `@next` versions to npm routinely as we merge PRs (as they do now)
2. Also publish `@latest` versions to npm when we release

These two processes have to exist in the same workflow, because only one action can be designated as a 'trusted publisher' for an npm package.

7804626 is my current best-guess at how this could work:

- Leaves the existing `@next` publishing workflow as-is
- Adds a second job that is run only if a new release is published, and publishes to `@latest`

Once we settle on an approach, I will update `publish_design_tokens.yaml`.

### DIFF
--- a/.github/workflows/publish_react_component_library.yaml
+++ b/.github/workflows/publish_react_component_library.yaml
@@ -106,4 +106,3 @@ jobs:
       - name: Publish to npm @latest
         run: npm publish --tag latest
         working-directory: ./packages/react-components
-        


ty2k (changes requested):
One minor change request to add the newline back. Otherwise, I think this is a great place to start with the new npm OIDC flow. There's probably no great way to test this without merging and trying it out. The existing action is already broken due to the classic npm token expiry and deprecation, so you might as well merge this and just roll forward with any fixes for problems we find. :+1: :+1:

PR #555, opened 2025-11-20
Add repository object to package.json

mkernohanbc:
This PR attempts to resolve the error that is [causing the npm publish job to fail](https://github.com/bcgov/design-system/actions/runs/19549499927/job/55976941599). 

db59f6f adds a `repository` object to `package.json` that specifies the source repo and directory, so that npm can establish provenance.

This change will also need to be mirrored in the design tokens `package.json` once we validate that it works.

### DIFF
--- a/packages/react-components/package.json
+++ b/packages/react-components/package.json
@@ -3,7 +3,7 @@
   "version": "0.5.1",
   "repository": {
     "type": "git",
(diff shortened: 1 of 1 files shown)


ty2k (approved):
Looks good to me, one note on the `repository.url` piece below.

PR #565, opened 2025-11-27
Fix release tag matching in GHA to publish to npm

mkernohanbc:
This PR changes the criterion that [this GHA](https://github.com/bcgov/design-system/blob/main/.github/workflows/publish_react_component_library.yaml) uses to trigger a publish to npm.

The workflow now triggers when it detects a release tag name containing `bcgov/design-system-react-components`.


ty2k (approved):
:+1: :+1:

PR #570, opened 2025-11-28
remove old publish script from workflow and properly target dist folder

mkernohanbc:
This PR attempts to fix [this job failure](https://github.com/bcgov/design-system/actions/runs/19770090338/job/56652178421#step:11:40) in the GHA that publishes new versions of the design tokens library.

It removes an old script (`run-publish-npm-package`) from the workflow, and updates the publish steps in both the `next` and `latest` jobs to target the correct subfolder (`/dist`).


ty2k (approved):
✅

PR #575, opened 2025-12-04
Add NumberField component

mkernohanbc:
This PR implements a new `NumberField` component, using [RAC NumberField](https://react-spectrum.adobe.com/react-aria/NumberField.html) as its base.

Examples are included in the Vite kitchen sink app, along with examples and tech docs in Storybook:

Screenshot 2025-12-04 at 2 28 31 PM

A `NumberField` is composed from RAC's Group and Input components and the existing BCDS Button component. A numerical value can be entered via increment/decrement buttons, keyboard or using the scroll-wheel.

The `formatOptions` prop enables developers to define specific unit types and formatting (eg. currency or distance). Additionally, it also includes a `unitLabel` prop that renders an additional text label next to the input, intended for use with custom units not supported in `Intl.NumberFormat()`.

It supports the same form validation/submission and error-handling methods (via the `isInvalid` and `errorMessage` props) as our other input components. Developers can set `minValue` and/or `maxValue` to constrain the range of acceptable values. The `step` prop can also be used to set the increment 'tick'.

### DIFF
--- a/packages/react-components/src/components/NumberField/NumberField.tsx
+++ b/packages/react-components/src/components/NumberField/NumberField.tsx
@@ -26,14 +26,12 @@ interface NumberFieldProps extends ReactAriaNumberFieldProps {
   description?: string;
   /* Used for data validation and error handling */
   errorMessage?: string | ((validation: ValidationResult) => string);
-  unitLabel?: string;
 }
 
 export default function NumberField({
   size = "medium",
   label,
   description,
-  unitLabel,
   errorMessage,
   ...props
 }: NumberFieldProps) {
@@ -75,14 +73,6 @@ export default function NumberField({
             </div>
             <div className="bcds-react-aria-NumberField--InputContainer">
               <Input className="bcds-react-aria-NumberField--Input" />
-              {unitLabel && (
-                <span
-                  className="bcds-react-aria-NumberField--UnitLabel"
-                  slot="label"
-                >
-                  {unitLabel}
-                </span>
-              )}
               {isInvalid && <SvgExclamationIcon />}
             </div>
             <div className="bcds-react-aria-NumberField--Button">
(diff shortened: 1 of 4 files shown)


mkernohanbc:
Leaving in draft pending unit tests

mkernohanbc:
> Check to see if the disabled text in the number value and units are using the disabled type token

Fixed in [bab4634](https://github.com/bcgov/design-system/pull/575/commits/bab46349e593844da06b7f05be48bd529145764b)

packages/react-components/src/components/NumberField/NumberField.tsx
```diff
@@ -0,0 +1,118 @@
...
+import Button from "../Button";
+import SvgExclamationIcon from "../Icons/SvgExclamationIcon";
+import SvgMinusIcon from "../Icons/SvgMinusIcon";
+import SvgPlusIcon from "../Icons/SvgPlusIcon";
+import Text from "../Text";
+
+import "./NumberField.css";
+
+interface NumberFieldProps extends ReactAriaNumberFieldProps {
+  size?: "small" | "medium";
```
ty2k: Can we please get comments on these? See [Button as an example](https://github.com/bcgov/design-system/blob/main/packages/react-components/src/components/Button/Button.tsx#L8).

packages/react-components/src/stories/NumberField.mdx
```diff
@@ -0,0 +1,128 @@
...
+
+By default, input fields are marked as optional. Use the `isRequired` prop to mark an input as required, and display a text label:
+
+<Canvas of={NumberFieldStories.Required} />
+
+### Controlling the value
+
+By default, a NumberField's value is uncontrolled.
+
+You can set its initial using the `defaultValue`. You can also control the NumberField using your preferred state management library, and set or its value via the `value` prop.
```
ty2k: Typo in `and set or its value `.

PR #581, opened 2025-12-15
Add ToggleButton and ToggleButtonGroup components

mkernohanbc:
This PR adds two new components: `ToggleButton` and `ToggleButtonGroup` (implemented based on [RAC components of the same names](https://react-spectrum.adobe.com/react-aria/ToggleButton.html).)

A `ToggleButton` can be toggled between binary states (on/off):

Inactive:
Screenshot 2025-01-20 at 16 50 09

Active:
Screenshot 2025-01-20 at 16 50 42

Multiple buttons can be rendered inside a `ToggleButtonGroup` to create more complex UIs (for example, choosing between mutually-exclusive options):

Screenshot 2025-01-20 at 16 51 36

Docs and examples are provided in Storybook.

### DIFF
--- a/packages/react-components/src/components/ToggleButton/ToggleButton.css
+++ b/packages/react-components/src/components/ToggleButton/ToggleButton.css
@@ -60,6 +60,12 @@
   border: var(--layout-border-width-small) solid
     var(--surface-color-secondary-button-default);
   box-shadow: var(--surface-shadow-small);
+  /* Animate transition on select or deselect */
+  transition: 200ms ease;
+
+  @media (prefers-reduced-motion: reduce) {
+    transition: none;
+  }
 }
 
 /* Hover */
@@ -87,4 +93,10 @@
 .bcds-react-aria-ToggleButton[data-disabled] {
   color: var(--typography-color-disabled);
   cursor: not-allowed;
+  /* Animate transition on disable or enable */
+  transition: 200ms ease;
+
+  @media (prefers-reduced-motion: reduce) {
+    transition: none;
+  }
 }
--- a/packages/react-components/src/components/ToggleButtonGroup/ToggleButtonGroup.css
+++ b/packages/react-components/src/components/ToggleButtonGroup/ToggleButtonGroup.css
@@ -27,39 +27,3 @@
   font: var(--typography-regular-small-body);
   color: var(--typography-color-secondary);
 }
-
-/* Selection indicator */
-.bcds-react-aria-ToggleButton:has(
-  .bcds-react-aria-ToggleButton--SelectionIndicator
-) {
-  position: relative;
-  transition: 200ms ease;
-
-  @media (prefers-reduced-motion: reduce) {
-    transition: none;
-  }
-}
-
-.bcds-react-aria-ToggleButtonGroup
-  > .bcds-react-aria-ToggleButton
-  > .bcds-react-aria-ToggleButton--SelectionIndicator[data-selected] {
-  position: absolute;
-  z-index: 0;
-  top: 0;
-  left: 0;
-  right: 0;
-  bottom: 0;
-  background-color: var(--surface-color-secondary-button-default);
-  color: var(--typography-color-primary);
-  border: var(--layout-border-width-small) solid
-    var(--surface-color-secondary-button-default);
-  box-shadow: var(--surface-shadow-small);
-  border-radius: var(--layout-border-radius-medium);
-  box-shadow: var(--surface-shadow-small);
-  pointer-events: none;
-}
-
-.bcds-react-aria-ToggleButtonGroup > .bcds-react-aria-ToggleButton > span {
-  position: relative;
-  z-index: 1;
-}
--- a/packages/react-components/src/components/ToggleButtonGroup/ToggleButtonGroup.tsx
+++ b/packages/react-components/src/components/ToggleButtonGroup/ToggleButtonGroup.tsx
@@ -2,16 +2,11 @@ import {
   ToggleButtonGroup as ReactAriaToggleButtonGroup,
   ToggleButtonGroupProps as ReactAriaToggleButtonGroupProps,
   Text,
-  SelectionIndicator,
-  composeRenderProps,
 } from "react-aria-components";
-import { Children, cloneElement, isValidElement, ReactElement } from "react";
-import { ToggleButtonProps } from "../ToggleButton/ToggleButton";
 
 import "./ToggleButtonGroup.css";
 
-export interface ToggleButtonGroupProps
-  extends ReactAriaToggleButtonGroupProps {
+export interface ToggleButtonGroupProps extends ReactAriaToggleButtonGroupProps {
   label?: string;
   size?: "small" | "medium";
 }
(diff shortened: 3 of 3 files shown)


mkernohanbc:
> The active/clicked state of the toggle button is using the wrong outline color (right now its orange).

This is weird, I'm seeing the correct blue focus ring:

Screenshot 2026-01-05 at 1 52 25 PM

There's no usage of that orange anywhere in the CSS, and it doesn't look like a browser default. What were the steps to get to this?

Philip-Cheung:
> > The active/clicked state of the toggle button is using the wrong outline color (right now its orange).
> 
> This is weird, I'm seeing the correct blue focus ring:
> 
> Screenshot 2026-01-05 at 1 52 25 PM
> There's no usage of that orange anywhere in the CSS, and it doesn't look like a browser default. What were the steps to get to this?
Screenshot 2026-01-06 at 9 38 36 AM

I found the orange ring via an LLM IDE's built-in browser I'm using (cursor).

packages/react-components/src/stories/ToggleButtonGroup.stories.tsx
```diff
@@ -0,0 +1,182 @@
...
+  },
+} satisfies Meta<typeof ToggleButtonGroup>;
+
+export default meta;
+type Story = StoryObj<typeof meta>;
+
+export const ToggleButtonGroupTemplate: Story = {
+  args: {
+    children: [
+      <ToggleButton id="1">Button 1</ToggleButton>,
```
ty2k: All of the `children` arrays of `ToggleButton`s in this file need `key` props to avoid console errors.

packages/react-components/src/stories/ToggleButtonGroup.stories.tsx
```diff
@@ -0,0 +1,182 @@
...
+  render: ({ ...args }: ToggleButtonGroupProps) => (
+    <ToggleButtonGroup {...args} />
+  ),
+};
+
+export const HorizontalToggleButtonGroup: Story = {
+  ...ToggleButtonGroupTemplate,
+  args: {
+    orientation: "horizontal",
+    children: [
```
ty2k: If you want to re-use `args` from the template story, you can do this to avoid the copy/paste repetition:

```tsx
export const HorizontalToggleButtonGroup: Story = {
  ...ToggleButtonGroupTemplate,
  args: {
    ...ToggleButtonGroupTemplate.args,
  },
};
```

... and then customize anything that needs customizing after that `...ToggleButtonGroupTemplate.args` spread.

packages/react-components/src/components/ToggleButtonGroup/ToggleButtonGroup.tsx
```diff
@@ -0,0 +1,65 @@
...
+          {label}
+        </Text>
+      )}
+      <ReactAriaToggleButtonGroup
+        className={`bcds-react-aria-ToggleButtonGroup ${
+          size === "small" ? "small" : "medium"
+        }`}
+        {...props}
+      >
+        {ToggleButtonChildren as React.ReactNode}
```
ty2k: Got a link to the RAC example that does this? This works, but I can't figure out why this isn't just a `{children}` slot.

mkernohanbc: This originates from the need to render a `<SelectionIndicator>` element for each child item in order to be able to animate the transition between states (see lines 25-46): https://react-aria.adobe.com/ToggleButtonGroup#animation

I suspect that ```{ToggleButtonChildren as React.ReactNode}``` was originally in response to a type error, but there doesn't appear to be an error when I remove it ([3c64e95](https://github.com/bcgov/design-system/pull/581/commits/3c64e95b611aed946b556fd73a0353332a8ed835)) so :shrug:

packages/react-components/src/components/ToggleButton/ToggleButton.css
```diff
@@ -0,0 +1,85 @@
+.bcds-react-aria-ToggleButton {
+  display: inline-flex;
+  flex-direction: row;
+  box-sizing: border-box;
+  gap: var(--layout-margin-xsmall);
+  align-items: center;
+  justify-content: center;
+  padding: var(--layout-padding-hair) var(--layout-padding-small);
+  color: var(--typography-color-secondary);
+  background-color: var(--surface-color-secondary-button-disabled);
```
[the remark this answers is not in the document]
mkernohanbc: I think in this context we're treating "disabled" as a synonym for "inactive". I'll add this to the list of things to consider for a future expansion of the tokens schema.

packages/react-components/src/components/ToggleButtonGroup/ToggleButtonGroup.css
```diff
@@ -0,0 +1,68 @@
+.bcds-react-aria-ToggleButtonGroup {
+  display: flex;
+  flex: 1 1 0;
+  flex-wrap: wrap;
```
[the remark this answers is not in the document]
mkernohanbc: @Philip-Cheung as discussed, I've been lost in the weeds trying to figure out the least brittle styling logic. What's currently on this branch (as of [5073112](https://github.com/bcgov/design-system/pull/581/commits/50731124cedec1e835a2c8d1b3619f2d72ac9e4f)) represents my best guess at a flex-based approach that sacrifices perfect symmetry for handling most edge and corner cases (very long button labels, large numbers of buttons, small viewport sizes, logical behaviour in horizontal/vertical orientations, etc.)

I'm going to circle back on this next week, but feel free to pull down the branch and have a play around in the meantime.

packages/react-components/src/components/ToggleButtonGroup/ToggleButtonGroup.tsx
```diff
@@ -0,0 +1,65 @@
...
+  const ToggleButtonChildren = Children.map(
+    children as React.ReactNode,
+    (child) => {
+      if (isValidElement(child)) {
+        return cloneElement(child as ReactElement<ToggleButtonProps>, {
+          children: composeRenderProps(
+            (child as ReactElement<ToggleButtonProps>).props.children,
+            (children) => (
+              <>
+                <SelectionIndicator
```
ty2k: I wonder: with the new animation, you're not really using the `SelectionIndicator` as a visible element that moves between choices anymore (like in [this RAC docs example](https://react-aria.adobe.com/selection#animated-selectionindicator)). Does it make sense to use it at all now? In my mind, with the new fade-in/fade-out animation, you could really simplify the React component markup and the CSS by just styling the selected button based on `[data-selected]`. 

If you're going with the current animation style, I can only see pros from getting rid of the `SelectionIndicator`:

- In your React component, you can get rid of all the `ToggleButtonChildren` logic and just stick `{children}` straight in that slot without the extra composition step. The means you can get rid of your `<span>` wrapper around the children (see comment in `ToggleButton.css`), allowing that `gap` property to work as expected.
- In your CSS, you can avoid the `z-index` and positioning stuff, which can cause visual bugs with different stacking contexts (think: a modal that has a `z-index` of 10 holds a `ToggleButtonGroup` where elements have a `z-index` of 0 or 1, placing them visually behind the modal).

Just less code and less DOM nodes on the page to think about in general.

ty2k (changes requested):
This looks really awesome now. I think it can be refactored slightly to make it way simpler by taking out the `SelectionIndicator`.

PR #593, opened 2025-12-17
Apply focus styling more consistently

mkernohanbc:
This PR makes changes to 5 components to handle focus states more consistently. Affected components are:

- Checkbox
- Select
- Tag
- TextArea
- TextField

The key change is to the conditions for applying the offset focus ring, like this:

Screenshot 2025-12-17 at 2 53 47 PM

In these (mostly older) components, we previously used the RAC data attribute `[data-focused]` to apply this style. This triggers whenever the element is in focus, whether via keyboard or via the user clicking into it, etc.

This change updates these components to instead use the attribute `[data-focus-visible]`, which applies more narrowly **only** when the element is keyboard-focused. This is the approach we have adopted for most of our more recent components.

TextField in non-keyboard focus (`[data-focused]`):
Screenshot 2025-12-17 at 2 53 08 PM

TextField in keyboard focus (`[data-focus-visible]`):
Screenshot 2025-12-17 at 2 52 59 PM

This change preserves the functionality of clearly indicating where focus is for users navigating via keyboard, but reduces visual load for other users (letting hover, border and cursor visual cues handle the task.)

### DIFF
--- a/packages/react-components/src/components/Checkbox/Checkbox.css
+++ b/packages/react-components/src/components/Checkbox/Checkbox.css
@@ -32,7 +32,7 @@
 }
 
 /* Focus */
-.bcds-react-aria-Checkbox[data-focused] > .checkbox {
+.bcds-react-aria-Checkbox[data-focus-visible] > .checkbox {
   outline: solid var(--layout-border-width-medium)
     var(--surface-color-border-active);
   outline-offset: var(--layout-margin-hair);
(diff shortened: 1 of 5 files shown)


mkernohanbc:
@Philip-Cheung let's touch base in the new year, and confirm we're being consistent (in code and Figma, and from component to component.)

Philip-Cheung (approved):

mkernohanbc:
Holding on merging this PR, will use it as a test case for publish workflow changes in #677.

PR #601, opened 2025-12-28
Split React component library build concerns, migrate ESLint to v9

ty2k:
This PR moves all of the React component library's `devDependencies` to latest where possible. This involves some refactoring for moving `@rollup/plugin-typescript` to v12 and `eslint` to v9.

## Splitting build concerns

As part of the move to `@rollup/plugin-typescript` at v12, the Rollup build process had to be refactored.

Because we supply both ESM and CJS versions of the React components, we have always had our TypeScript config's `declarationDir` outside of the individual ESM and CSJ output directories specified in Rollup. This is no longer allowed as part of a breaking change between v11 and v12. See [this issue](https://github.com/rollup/plugins/issues/1772).

In refactoring how the Rollup process works, I made it modular by breaking apart the bundling steps. Now, instead of Rollup doing the entire bundling process itself, the steps look like: 

1. Using Rollup, bundle the ESM and CJS code (this is the `rollup.config.js` file used in the new `build:js` script in `package.json`)
2. Using TypeScript's `tsc`, build the type declarations for the project (see `build:types`)
3. Using Rollup, bundle the types into a single file `index.d.ts` for distribution (see `rollup.dts.config` used by `build:dts`)

To reflect that Rollup isn't doing all the building/bundling itself, I've renamed the aggregate `rollup` script to `build`, and broken out the action of deleting the `dist` directory into a separate `clean` script. The `build` script runs `clean`, `build:js`, `build:types`, and `build:dts` in that order.

Separate TypeScript configuration files are shipped for the `build:js` and `build:types` steps. These extend the base config from `tsconfig.json`, but with changes to turn declarations on or off. 

The files that get published are specified in the `files` array of `package.json` (see [npm publish documentation on files included in package](https://docs.npmjs.com/cli/v8/commands/npm-publish?v=true#files-included-in-package)). The individual TypeScript type declaration files will no longer be published along with the aggregated one. Users shouldn't notice a difference here - the `types` declaration in `package.json` should point IDEs to the aggregated file for type declarations.

These changes should make it easier to migrate to [Rolldown](https://rolldown.rs/guide/introduction) in the future. Vite will be defaulting to Rolldown in v8, so that would be a good thing to do at the same time if possible.

## ESLint v9 migration

The name and structure of the ESLint configuration file had to change to match the new standard in v9. The new format is less magical, with explicit imports and references to them rather than passing references by string. The `globalIgnores` section of the new ESLint configuration now explicitly excludes the different build directories so the lint script should run and pass no matter what's happening on the developer's machine.

I made one small change to the typing of the Header component in e858662709f261e6a5c2cf4633b9f8dd9082f0e4 to explicitly cast the user-supplied `logoLinkElement` as type `React.ReactElement<any>`. The previous type `JSX.Element` implies that only elements created from JSX syntax are allowed, though we really want this to be anything the user wants (a `<Link>` from something like React Router or Next.js, a standard `<a>` link, or anything else). I used `eslint-disable-next-line` to allow the `any` usage to pass here - we don't generally want to use it, but this use of `cloneElement()` in the Header is a special case.

Right now, the lint process completes without error. 

## Dependencies

A handful of dependencies are not updated to their latest versions because some Storybook dependencies don't yet support the latest. These are:

- `@vitest/browser-playwright` (using 4.0.14, 4.0.16 is current lastest)
- `@vitest/coverage-v8` (using 4.0.14, 4.0.16 is current latest)
- `vitest` (4.0.14, 4.0.16 is current latest)

Some new dependencies are added for ESLint v9:

- `globals`
- `typescript-eslint`

Some old dependencies from Jest are pruned in this PR:

- `identity-obj-proxy` ("an identity object using ES6 proxies", Vitest handles this functionality automatically)
- `ts-node` (an alternate TypeScript runtime/REPL that went in at the same time as Jest which is not used by anything else in the package)

## Documentation updates

The README is updated to include linting and build details. Testing instructions for local builds are included.

## Testing

I've tested that every script in `package.json` runs and completes without error and with the build output I expect (where applicable). Both Dockerfiles build and run. All test suites are passing.

To test the React components library, I've used `npm link` locally in new projects. See updated README for documentation on this process.

I haven't tested the GitHub Actions workflows directly, but running the individual scripts in each workflow works as expected.


mkernohanbc (approved):
This is great — thanks so much @ty2k!

PR #602, opened 2026-01-05
Change how npm registry URL is set

mkernohanbc:
This PR makes a small change to simplify the GHA used to publish `design-system-react-components` to npm. It mirrors one of the changes made in #592. 

The `registry-url` value is now set via `setup-node`, instead of in a discrete step later in the workflow.


ty2k (approved):
:+1: :+1:

PR #611, opened 2026-01-13
Add RAC docs cards to Storybook pages

mkernohanbc:
This PR aims to improve wayfinding to the full API documentation for components based on React Aria primitives.

It adds an `InlineAlert` to the top of each such component's docs in Storybook, to clearly indicate at a glance:

- That the component is based on a RAC foundation
- Which RAC components and subcomponents it uses
- Where the user can find the full API reference

Screenshot 2026-01-13 at 3 16 29 PM

This PR also implements a couple of new icon components to support this use-case (`SvgReactAriaIcon` and `SvgUpRightFromSquareIcon`.)

### DIFF
--- a/packages/react-components/src/stories/AccordionGroup.mdx
+++ b/packages/react-components/src/stories/AccordionGroup.mdx
@@ -40,7 +40,7 @@ import {
       size="small"
       buttonVariant="secondary"
       href="https://react-aria.adobe.com/Disclosure"
-      target="blank"
+      target="_blank"
       iconRight={[<SvgUpRightFromSquareIcon size={16} />]}
     >
       Disclosure docs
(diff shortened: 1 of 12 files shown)


packages/react-components/src/stories/AccordionGroup.mdx
```diff
@@ -23,20 +29,47 @@ import { AccordionGroup } from "@/components";
...
+  customIcon={[<SvgReactAriaIcon />]}
+  title="This is a React Aria component"
+  description="It is based on the RAC Disclosure and DisclosureGroup components. Consult the React Aria docs to understand its full API."
+  buttons={[
+    <Link
+      isButton
+      size="small"
+      buttonVariant="secondary"
+      href="https://react-aria.adobe.com/Disclosure"
+      target="blank"
```
ty2k: All of these should have [`target`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/a#target) set to `_blank` with the leading underscore. When you use `blank`, the first time the user clicks the link, they end up opening the link in a new context named `blank` (this opens in a new tab in Chrome). The second time a user clicks a link with target `blank`, it actually opens in that same tab, overwriting what's in the context of `blank`.

ty2k (approved):
🚀 Looks great!

PR #618, opened 2026-01-26
Add DatePicker and Calendar components

mkernohanbc:
This PR adds a new DatePicker component, based on [RAC DatePicker](https://react-aria.adobe.com/DatePicker). A date picker allows the user to enter a date or date/time value by typing, or by incrementing day/month/year/hour/minute/second segments with the keyboard:

Screenshot 2026-01-26 at 1 48 58 PM

It also uses `Button`, `Popover` and a new `Calendar` component (based on [RAC Calendar](https://react-aria.adobe.com/Calendar) to let the user select a date from a calendar: 

Screenshot 2026-01-26 at 1 46 15 PM

**Note**: Calendar is currently consumed exclusively by DatePicker, and is not yet fully exposed/documented. This implementation anticipates future use-cases outside the context of DatePicker.

DatePicker is patterned similarly to other BCDS input components, and supports the standard set of states and props:

- Required/optional
- Disabled and read-only states
- Validation and error handling via `FieldError` and the `errorMessage` prop

Developers can constrain acceptable date ranges via `minValue` and `maxValue` props:

Screenshot 2026-01-26 at 1 52 38 PM

They can also selectively remove dates by passing a callback to `isDateUnavailable`:

Screenshot 2026-01-26 at 1 52 27 PM

### DIFF
--- a/packages/react-components/src/components/DatePicker/DatePicker.tsx
+++ b/packages/react-components/src/components/DatePicker/DatePicker.tsx
@@ -8,9 +8,15 @@ import {
   Group,
   Label,
   Popover,
+  Separator,
   Text,
   ValidationResult,
+  I18nProvider,
 } from "react-aria-components";
+
+import { useMemo } from "react";
+import { useDisplayNames } from "@react-aria/datepicker";
+import { useDateFormatter } from "@react-aria/i18n";
 
 import "./DatePicker.css";
 import Button from "../Button";
@@ -30,22 +36,67 @@ export interface DatePickerProps<
   label?: string;
   /* Sets optional description text below input */
   description?: string;
+  /* Show date format in the description */
+  showFormatHelpText?: boolean;
   /* Used for data validation and error handling */
   errorMessage?: string | ((validation: ValidationResult) => string);
+  /* When true, uses the browser locale for date formatting and internationalization */
+  isBrowserLocaleUsed?: boolean;
 }
 
 export type { DateValue };
+
+/* Generate format helper text, adapted from React Spectrum */
+function useFormatHelpText(
+  props: Pick<DatePickerProps<DateValue>, "showFormatHelpText">,
+): React.ReactNode {
+  const formatter = useDateFormatter({ dateStyle: "short" });
+  const displayNames = useDisplayNames();
+  return useMemo(() => {
+    if (props.showFormatHelpText) {
+      return (
+        <>
+          <span>Format: </span>
+          {formatter.formatToParts(new Date()).map((s, i) => {
+            if (
+              s.type === "literal" ||
+              s.type === "unknown" ||
+              (s.type as string) === "yearName"
+            ) {
+              return <span key={i}>{` ${s.value} `}</span>;
+            }
+
+            const type = (s.type as string) === "relatedYear" ? "year" : s.type;
+            return (
+              <span
+                key={i}
+                className="bcds-react-aria-DatePicker--FormatHelpText"
+              >
+                {displayNames.of(type)}
+              </span>
+            );
+          })}
+        </>
+      );
+    }
+
+    return "";
+  }, [props.showFormatHelpText, formatter, displayNames]);
+}
 
 export default function DatePicker<T extends DateValue>({
   size = "medium",
   label,
   description,
   errorMessage,
+  showFormatHelpText = true,
   isRequired = false,
   isCalendarDisabled = false,
+  isBrowserLocaleUsed = false,
   ...props
 }: DatePickerProps<T>) {
-  return (
+  const formatHelpText = useFormatHelpText({ showFormatHelpText });
+  const datePicker = (
     <ReactAriaDatePicker
       className={`bcds-react-aria-DatePicker ${size}`}
       {...props}
@@ -85,13 +136,18 @@ export default function DatePicker<T extends DateValue>({
               </>
             )}
           </Group>
-
-          {description && (
+          {(description || showFormatHelpText) && (
             <Text
               slot="description"
               className={`bcds-react-aria-DatePicker--Description`}
             >
-              {description}
+              {showFormatHelpText && (
+                <>
+                  {formatHelpText}
+                  <Separator className="bcds-react-aria-DatePicker--Separator" />
+                </>
+              )}
+              <div>{description && description}</div>
             </Text>
           )}
           <FieldError className="bcds-react-aria-DatePicker--Error">
@@ -100,5 +156,11 @@ export default function DatePicker<T extends DateValue>({
         </>
       )}
     </ReactAriaDatePicker>
+  );
+
+  return isBrowserLocaleUsed ? (
+    datePicker
+  ) : (
+    <I18nProvider locale={"en-CA"}>{datePicker}</I18nProvider>
   );
 }
--- a/packages/react-components/src/components/DatePicker/DatePicker.css
+++ b/packages/react-components/src/components/DatePicker/DatePicker.css
@@ -76,6 +76,18 @@
   min-width: var(--icons-size-medium);
   height: var(--icons-size-medium);
 }
+/* Format helper text */
+.bcds-react-aria-DatePicker--FormatHelpText {
+  unicode-bidi: embed;
+  direction: ltr;
+}
+/* Separator */
+.bcds-react-aria-DatePicker--Separator {
+  margin: var(--layout-margin-small) var(--layout-margin-none);
+  height: var(--layout-border-width-small);
+  border: none;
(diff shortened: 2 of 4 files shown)


mkernohanbc:
> I was pressing "tab" a bunch of times to check out the focus state and when I got to the selected date here, the date's font becomes white (and is hard to read).

This issue is resolved in [3e0caee](https://github.com/bcgov/design-system/pull/618/commits/3e0caeed3faab00b4d6138037729be12b267277a)

packages/react-components/src/components/Calendar/Calendar.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+import Button from "../Button";
+import Heading from "../Heading";
+import SvgChevronLeftIcon from "../Icons/SvgChevronLeftIcon";
+import SvgChevronRightIcon from "../Icons/SvgChevronRightIcon";
+
+export interface CalendarProps<
+  T extends DateValue,
+> extends ReactAriaCalendarProps<T> {
+  isHeaderHidden?: boolean;
```
ty2k: It looks like this prop isn't being used right now with the code in this pull request. I would err on the side of leaving this out and only adding it if you have a valid use case for it somewhere down the line. It's an easy addition if you need it later because your default case is to include the header controls.

packages/react-components/src/components/DatePicker/DatePicker.tsx
```diff
@@ -0,0 +1,99 @@
...
+import "./DatePicker.css";
+import Button from "../Button";
+import Calendar from "../Calendar";
+import Dialog from "../Dialog";
+import SvgCalendarIcon from "../Icons/SvgCalendarIcon";
+import SvgExclamationIcon from "../Icons/SvgExclamationIcon";
+
+export interface DatePickerProps<
+  T extends DateValue,
+> extends ReactAriaDatePickerProps<T> {
```
ty2k: Can we get comments on these props please? Ex:

```ts
  /** Height of the control, defaults to medium. */
  size?: "small" | "medium";
```

packages/react-components/src/components/DatePicker/DatePicker.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+export type { DateValue };
+
+export default function DatePicker<T extends DateValue>({
+  size = "medium",
+  label,
+  description,
+  errorMessage,
+  isRequired = false,
+  isCalendarDisabled = false,
```
ty2k: If a `DatePicker` has the calendar disabled, isn't it just a `DateField`? If so, it might make more sense to just ship a `DateField` and leave this logic out of the `DatePicker` to better match the RAC API. There might be some use case I can't think of that you have in mind, though!

ty2k (changes requested):
This is really excellent and basically ready to go, IMO. Awesome job. That being said, here's an essay lol.

I worry that the RAC style of showing the date format only as placeholder text means a big loss of accessibility. This is compounded because this component is internationalized by default, which is complicated in a government context like ours. The internationalization aspect leaves us without a good way to proactively label the date format beyond the placeholder text if we follow the default pattern.

I tested this code out on my personal machine where the browser locales are all set to `en-US`. I assume this might be a system default since it's not the type of thing I would proactively set or change. On our work machines with enterprise administration, systems get defaulted to `en-CA`. The `en-US` setting results in a really heinous date format for the component:

en-US date format before entry

Users can work with an unfamiliar format, but when the only affordance you have for the date format is placeholder text, it means you have to either open the calendar (if available) or completely clear the text field to make sure you've entered the right thing:

en-US date format filled in

Is this January 2 or February 1?

Without having the information in the label or description, we increase the cognitive load for someone entering the form data themselves, but we really screw things up for people who might be helping others. Imagine a scenario like helping an elderly relative fill out the form over the phone and your date format happens to be different than theirs, for example.

IMO, we might want to default to passing `en-CA` to an `I18nProvider` wrapping the component, while providing a fallback if folks really want the "browser default internationalization" version. I feel like the `en-CA` default will be what we want in the vast majority of cases.

By defaulting to `en-CA`, we can always know the date format and the developer can add it right to the label for whatever date format they're using. Ideally this wouldn't even be a manual thing for them to do, but I can't think of a way to do it automatically.

Image

The way I propose doing this is with a boolean prop like `isBrowserLocaleUsed` that defaults to `false`. I'll put a code block with the full component at the bottom of this post. With this idea, internationalized pages/apps could always use the RAC style browser default internationalization by passing the `isBrowserLocaleUsed` prop. Pages that are specifically designed in a given language/locale could override whatever the browser has by passing the `isBrowserLocaleUsed` prop and then using the RAC `I18nProvider` wrapper with a `locale` of their choice.

I realize "you are not the user" so this might be me making a mountain out of mole hill. But I think it would be best to make the best accessible version of this the default while providing off ramps for the edge cases that need it. You could ask the Analytics team what browser locale stats look like for Gov.bc.ca right now to better guesstimate the scope of this issue.

Curious to see what you think on this one, as it probably never came up when testing on your gov machine.

Here's the edits I made with a `isBrowserLocaleUsed` prop:

```tsx
import {
  DatePicker as ReactAriaDatePicker,
  DatePickerProps as ReactAriaDatePickerProps,
  DateValue,
  DateInput,
  DateSegment,
  FieldError,
  Group,
  I18nProvider, // Import the internationalization provider
  Label,
  Popover,
  Text,
  ValidationResult,
} from "react-aria-components";

import "./DatePicker.css";
import Button from "../Button";
import Calendar from "../Calendar";
import Dialog from "../Dialog";
import SvgCalendarIcon from "../Icons/SvgCalendarIcon";
import SvgExclamationIcon from "../Icons/SvgExclamationIcon";

export interface DatePickerProps<
  T extends DateValue,
> extends ReactAriaDatePickerProps<T> {
  size?: "small" | "medium";
  isBrowserLocaleUsed?: boolean; // Add the prop for defaulting to i18n
  isCalendarDisabled?: boolean;
  label?: string;
  description?: string;
  errorMessage?: string | ((validation: ValidationResult) => string);
}

export type { DateValue };

export default function DatePicker<T extends DateValue>({
  size = "medium",
  label,
  description,
  errorMessage,
  isRequired = false,
  isCalendarDisabled = false,
  isBrowserLocaleUsed = false, // Default to false because want to use `en-CA` mostly
  ...props
}: DatePickerProps<T>) {
  const datePicker = (
    <ReactAriaDatePicker
      className={`bcds-react-aria-DatePicker ${size}`}
      {...props}
    >
      {({ isInvalid }) => (
        <>
          {label && (
            <Label className="bcds-react-aria-DatePicker--Label">
              {label}
              {isRequired && (
                <span className="bcds-react-aria-DatePicker--Label required">
                  (required)
                </span>
              )}
            </Label>
          )}
          <Group className="bcds-react-aria-DatePicker--Group">
            <DateInput className="bcds-react-aria-DatePicker--Input">
              {(segment) => (
                <DateSegment
                  className="bcds-react-aria-DatePicker--Segment"
                  segment={segment}
                />
              )}
            </DateInput>
            {isInvalid && <SvgExclamationIcon />}
            {!isCalendarDisabled && (
              <>
                <Button size="xsmall" variant="secondary" isIconButton>
                  <SvgCalendarIcon />
                </Button>
                <Popover>
                  <Dialog isCloseable={false}>
                    <Calendar />
                  </Dialog>
                </Popover>
              </>
            )}
          </Group>

          {description && (
            <Text
              slot="description"
              className={`bcds-react-aria-DatePicker--Description`}
            >
              {description}
            </Text>
          )}
          <FieldError className="bcds-react-aria-DatePicker--Error">
            {errorMessage}
          </FieldError>
        </>
      )}
    </ReactAriaDatePicker>
  );

  // The default state is to end up with the `I18nProvider` wrapper set to `en-CA`
  return isBrowserLocaleUsed ? (
    datePicker
  ) : (
    <I18nProvider locale={"en-CA"}>{datePicker}</I18nProvider>
  );
}
```

packages/react-components/src/components/Icons/SvgCalendarIcon/SvgCalendarIcon.tsx
```diff
@@ -0,0 +1,17 @@
+/* The component implements the Calendar icon from Font Awesome: https://fontawesome.com/icons/calendar */
+export default function SvgCalendarIcon({ id = "calendar-icon" }) {
```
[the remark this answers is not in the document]
mkernohanbc: Fixed in [0172038](https://github.com/bcgov/design-system/pull/618/commits/01720382a284fcc5f3c1c32d5c411ba39062450e)

PR #620, opened 2026-02-02
DatePicker: rework i18n implementation

mkernohanbc:
This PR makes changes to how `DatePicker` leverages React Aria's i18n featureset, based on [this suggestion](https://github.com/bcgov/design-system/pull/618#pullrequestreview-3734723347) from @ty2k:

- Adds a new ~`formatLabel`~ `showFormatHelpText` prop (defaults to `true`) that optionally displays the expected date format below the input, within the `description` slot. This is additional to, rather than overwriting, a description string
- Adds an `isBrowserLocaleUsed` prop (defaults to `false`) that enables developers to toggle between defaulting the date format to en-CA, or letting it set itself automatically based on the user's browser locale

When `isBrowserLocaleUsed = false`, the DatePicker renders inside an [`<I18nProvider>`](https://react-aria.adobe.com/I18nProvider) element with its locale set to `en-CA`. This overrides React Aria's built-in internationalization featureset to set a sane default for the majority of BCDS' expected use cases.

When `isBrowserLocaleUsed = true`, no `<I18nProvider>` is rendered, and the component will set its date format based on the user's browser locale. This is React Aria's default behaviour.

### DIFF
--- a/packages/react-components/src/components/DatePicker/DatePicker.tsx
+++ b/packages/react-components/src/components/DatePicker/DatePicker.tsx
@@ -47,11 +47,11 @@ export interface DatePickerProps<
 export type { DateValue };
 
 /* Generate format helper text, adapted from React Spectrum */
-export function useFormatHelpText(
+function useFormatHelpText(
   props: Pick<DatePickerProps<DateValue>, "showFormatHelpText">,
 ): React.ReactNode {
-  let formatter = useDateFormatter({ dateStyle: "short" });
-  let displayNames = useDisplayNames();
+  const formatter = useDateFormatter({ dateStyle: "short" });
+  const displayNames = useDisplayNames();
   return useMemo(() => {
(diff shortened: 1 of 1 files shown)


mkernohanbc:
[1ba0945](https://github.com/bcgov/design-system/pull/620/commits/1ba09453f30484185a43efaee19759095f7ee8b3) refactors how we implement the format label, based on React Spectrum's approach to solving this problem (see https://github.com/adobe/react-spectrum/discussions/9577#discussioncomment-15677993). 

`formatLabel` is renamed to `showFormatHelpText`, and a new `useFormatHelpText` function is added.

The advantages of adopting Spectrum's approach here: 

1. The strings for each part of the date value can now be automatically internationalized like the input value 
2. Simplifies the code a little, because `showFormatHelpText` doesn't change based on `isBrowserLocaleUsed` state

PR #630, opened 2026-02-05
Bug: DatePicker format label 

mkernohanbc:
This PR fixes a bug (introduced in #620) in the DatePicker component. 

The format label does not properly handle the case where:

- DatePicker is in its default state (inside an `i18nProvider` that sets its locale to `en-CA`) **and**
- The user's browser locale is set to something other than `en-CA`

In this case, the automatically-generated format label will reflect the browser locale, and thus conflict with the input value format:

Screenshot 2026-02-05 at 11 18 26 AM

3a61200fd3ae647153f7d45e0c06d33fce513eba fixes this by setting a default label for `en-CA` unless `isBrowserLocaleUsed` is passed:

Screenshot 2026-02-05 at 11 24 13 AM

This PR also includes a couple of housekeeping changes:

- 3dbb060934cf913dc0d7655409c07c2ae6fd319e adds a basic suite of unit tests for DatePicker
- 5a2c9ea6ea7092da083a0589a036d741f50e35be adds mocks of the DatePicker to the Vite kitchen sink app

### DIFF
--- a/packages/react-components/src/components/DatePicker/DatePicker.test.tsx
+++ b/packages/react-components/src/components/DatePicker/DatePicker.test.tsx
@@ -1,0 +1,62 @@
+import { describe, expect, test, vi } from "vitest";
+import { render, screen, fireEvent, within } from "@testing-library/react";
+import "@testing-library/jest-dom";
+import { CalendarDate } from "@internationalized/date";
+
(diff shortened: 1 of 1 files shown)


ty2k (approved):
Good fix!

DatePicker in en-CA and en-US

Issue #633, opened 2026-02-09
Button component has no min-width - should it?


ty2k:
### Component

Button

### Description

This isn't necessarily a bug, but maybe a decision that should be made: What happens if a `Button`'s contents are not wide enough to make the button at least as wide as it is tall? We have the `isIconButton` prop that forces a square (fixed width and fixed height), but we don't have a minimum width set on the different `size` options.

[Deque's current minimum touch target](https://dequeuniversity.com/rules/axe/4.11/target-size) is 24 x 24 CSS pixels (with caveats). Previous to adding the `xsmall` option for size, it wasn't really possible to generate anything that would create too small of a target, because we always had at least 24px of horizontal padding plus the button contents when using `small`.

Now with `xsmall` (16px of padding plus button contents), I find I can get touch target errors by lining up `xsmall` instances next to one another with no gap between them:

<img width="2032" height="1167" alt="Image" src="https://github.com/user-attachments/assets/139e9970-8f69-4d72-bf3c-cf2dfe72e1de" />

### Steps to reproduce

Do something like the following to generate touch target errors in Lighthouse:

```tsx
<div>
  <Button size="xsmall">q</Button>
  <Button size="xsmall">w</Button>
  <Button size="xsmall">e</Button>
  <Button size="xsmall">r</Button>
  <Button size="xsmall">t</Button>
  <Button size="xsmall">y</Button>
</div>
```

Force a minimum width equal to the minimum height of the `xsmall` size (24px) and there should be no errors:

```tsx
<div>
  <Button size="xsmall" style={{ minWidth: "24px" }}>q</Button>
  <Button size="xsmall" style={{ minWidth: "24px" }}>w</Button>
  <Button size="xsmall" style={{ minWidth: "24px" }}>e</Button>
  <Button size="xsmall" style={{ minWidth: "24px" }}>r</Button>
  <Button size="xsmall" style={{ minWidth: "24px" }}>t</Button>
  <Button size="xsmall" style={{ minWidth: "24px" }}>y</Button>
</div>
```

Note that adding a gap (like `gap: var(--layout-padding-xsmall);`) between the button instances will also create a satisfactory touch target size because the 24x24 target circles drawn by the accessibility rule won't have overlap potential (even with an empty button).

If we decided we wanted a button to always be at least as wide as it is tall, adding `min-width` alone might not be sufficient from a style perspective, because we probably would still want the contents of the `Button` centered.

### Technical information

[...]

PR #637, opened 2026-02-20
Button: add min-width and min-height

mkernohanbc:
This PR addresses the issue identified by @ty2k in #633. With the introduction of the `xsmall` button variant in 0.5.2, we inadvertently made it possible to violate [WCAG SC 2.5.8](https://www.w3.org/WAI/WCAG22/Understanding/target-size-minimum.html). An `xsmall` button with a single-character label fails to meet the minimum target size:

Screenshot 2026-02-18 at 11 33 05 AM

This change resolves this by setting `min-width` and `min-height` at the top level of the component. It also cleans up some inconsistency/duplication in various rules.

### DIFF
--- a/packages/react-components/src/components/Button/Button.css
+++ b/packages/react-components/src/components/Button/Button.css
@@ -1,11 +1,10 @@
 .bcds-react-aria-Button {
   display: inline-flex;
   align-items: center;
-  justify-content: space-around;
+  justify-content: center;
   gap: var(--layout-padding-small);
   min-height: 24px;
   min-width: 24px;
-  box-sizing: border-box;
   border: none;
   border-radius: var(--layout-border-radius-medium);
   cursor: pointer;


packages/react-components/src/components/Button/Button.css
```diff
@@ -1,10 +1,14 @@
...
-  border: none;
-  border-radius: var(--layout-border-radius-medium);
-  cursor: pointer;
   display: inline-flex;
   align-items: center;
+  justify-content: space-around;
   gap: var(--layout-padding-small);
+  min-height: 24px;
+  min-width: 24px;
+  box-sizing: border-box;
```
ty2k: `Button` always renders as an HTML `<button>` element, right? If so, this `box-sizing` rule can be removed, as `border-box` is the default for `<button>` elements.

ty2k (approved):
LGTM! 🚀

PR #642, opened 2026-03-02
DatePicker: fix separator logic

mkernohanbc:
This PR does two things:

1. 8b5565d629ced48dc13ea03202e727e161aa6937 fixes a bug in `DatePicker`, so that the separator between the format helper text and the description field is only shown when both elements are present 
2. b84fda5c870411e3e514eec8c74dfc75e1bbfa23 moves the styling of this separator element (based on [RA Separator](https://react-aria.adobe.com/Separator)) into a new component, so it can be consumed by other components as needed

Previous behaviour, for reference:
Screenshot 2026-03-02 at 8 58 17 AM

### DIFF
--- a/packages/react-components/src/components/DatePicker/DatePicker.css
+++ b/packages/react-components/src/components/DatePicker/DatePicker.css
@@ -81,13 +81,6 @@
   unicode-bidi: embed;
   direction: ltr;
 }
-/* Separator */
-.bcds-react-aria-DatePicker--Separator {
-  margin: var(--layout-margin-small) var(--layout-margin-none);
-  height: var(--layout-border-width-small);
-  border: none;
-  background-color: var(--surface-color-border-default);
-}
 
 /* Hover and focus */
 .bcds-react-aria-DatePicker--Group[data-hovered] {


packages/react-components/src/components/DatePicker/DatePicker.tsx
```diff
@@ -146,9 +146,9 @@ export default function DatePicker<T extends DateValue>({
                   {isBrowserLocaleUsed
                     ? formatHelpText
                     : "Format: year - month - day"}
-                  <Separator className="bcds-react-aria-DatePicker--Separator" />
```
ty2k: I think the matching styling from `DatePicker.css` can be removed too.

packages/react-components/src/stories/Separator.mdx
```diff
@@ -0,0 +1,57 @@
...
+<Subtitle>Visual divider for UI elements</Subtitle>
+
+<Source
+  code={`import { Separator } from "@bcgov/design-system-react-components";`}
+  language="typescript"
+/>
+
+## Usage and resources
+
+A separator provides visual separation between UI elements. It renders as a solid line.
```
ty2k: I'm guessing no RAC alert box for this one because it's too simple of a component, right?

ty2k (approved):
🏎️ 🏁

PR #643, opened 2026-03-04
Select: add support for multiple selections

### LINKED ISSUE #536: Add multi-select support to Select
As of RAC 1.13.0, the Select component now [supports multiple selections](https://react-spectrum.adobe.com/react-aria/Select.html#multiple-selection).

Further investigation needed to determine how we implement this in [our Select component](https://www2.gov.bc.ca/gov/content/digital/design-system/components/select).

mkernohanbc:
This PR makes significant (non-breaking) changes to Select, and minor changes to Tag. See the Vite kitchen sink app for examples. Changes are also documented in Storybook.

The impetus for this was the addition of [support for multi-select](https://react-aria.adobe.com/releases/v1-13-0) in React Aria v1.13.0. 

Summary of changes in this PR: 
- Select now supports toggling between single or multiple select via the `selectionMode` prop
- Added visual indicator for selected items in the ListBox
- Changed flex behaviour for Select so that it adjusts to its parent container and can grow vertically to show multiple rows
- `<SelectValue>` can now wrap over multiple lines, instead of being truncated with ellipsis
- Adjusted `small` style for Select to properly use `typography.regular.small.body`
- Added a new `xsmall` Tag variant
- Updated Tag to properly spread its props

In single-select mode, Select behaves as it currently does (showing the currently-selected value as a text string inside the Select button.) 

When `selectionMode` is set to `multiple`, it instead renders a `<TagGroup>` inside the button:

Screenshot 2026-03-04 at 11 22 46 AM

This means that selected items are:

- Visually separated
- De-selectable without re-opening the dropdown

Both the dropdown and the TagGroup are fully operable via keyboard:

https://github.com/user-attachments/assets/90d9ffe8-2dc4-4b1f-beef-10a3095b3a45

To prevent [an accessibility violation](https://dequeuniversity.com/rules/axe/4.11/nested-interactive?application=axeAPI) from having interactive tags nested inside a button, this PR uses [createPortal](https://react.dev/reference/react-dom/createPortal) to render the `<TagGroup>` outside the Select `<Button>`. This adds additional styling complexity, but achieves the desired visual structure accessibly. 

An alternative approach (f1d97d69b40475163899a4b4ed47fc5e3694a813) would be to render a `<Group>` instead of a `<Button>` when in multi-select mode. However, this approach:

- Breaks the sizing logic of the popover, because the `var(--trigger-width)` used to size the popover no longer works
- Forces a fundamental change in behaviour between Select modes, because in multi-select the dropdown could only be activated by pressing a small button, rather than by pressing anywhere on the Select object


Philip-Cheung:
Screenshot 2026-03-04 at 1 41 10 PM
The component crashed when I clicked on the "sections" control.

Philip-Cheung:
Screenshot 2026-03-04 at 1 45 34 PM

When I add selections into the input, does it expand by default as more items are added? Can I set a fixed width? I have a hunch a lot of folks are not going to like that behaviour as it messes up their forms/design

mkernohanbc:
> When I add selections into the input, does it expand by default as more items are added? Can I set a fixed width? I have a hunch a lot of folks are not going to like that behaviour as it messes up their forms/design

In this example, it's inside a flex container, so its width is constrained by the structure of the page. If there are too many items for the width, it can wrap to a new line cleanly:

Screenshot 2026-03-04 at 1 49 31 PM

packages/react-components/src/components/Select/Select.css
```diff
@@ -72,32 +120,90 @@
...
-  border: 1px solid var(--surface-color-border-default);
+  border: var(--layout-border-width-small) solid
+    var(--surface-color-border-default);
   border-radius: var(--layout-border-radius-medium);
   box-shadow: var(--surface-shadow-medium);
   box-sizing: border-box;
   overflow-y: auto;
   padding: var(--layout-padding-hair) var(--layout-padding-xsmall);
+  max-width: calc(100vw - (2 * var(--layout-padding-small)));
+  max-height: 50vh;
```
ty2k: What's the intent of setting `max-height` and `max-width` on the Popover? The Popover gets values added directly to it dynamically by RAC, so it's not a thing where I would think we need to setting anything like this, but there might be a use case I can't think of here. Note that there are cases where your `max-width` is going to get applied and `max-height` doesn't.

Image

packages/react-components/src/components/Select/Select.css
```diff
@@ -72,32 +120,90 @@
...
+  width: 100%;
+  min-width: 0;
+}
+.bcds-react-aria-Select--InputContainer > .bcds-react-aria-Select--Button {
+  grid-area: 1 / 1;
+  align-self: stretch;
+}
+.bcds-react-aria-Select--TagOverlayContainer {
+  grid-area: 1 / 1;
+  z-index: 1;
```
ty2k: Can't tell what this `z-index` does - if it's nothing, can we remove it? I'm always skeptical of it (see, for example, [Why we never use z-index](https://medium.com/finnovate-io/why-we-never-use-z-index-231034f1bc6f).

ty2k (changes requested):
Lots going on in this component now! I think the `createPortal()` stuff works nicely for using the TagGroup inside of the Select for multi-select. Overall, I can't really break this visually and it seems like the keyboard navigation is solid. Great stuff!

Couple thoughts beyond the specific line number comments below:

1. IMO, I think we need a visual indicator of some kind in multi-select mode when `overflow` is set to `scroll` and there is more information to be viewed. I see we're hiding the scrollbar here, but maybe there's an opportunity for something like the Gov.bc.ca gradient on the breadcrumbs, as an example? I think there's a risk of hiding information with this style in this general. I would personally opt for the `overflow: wrap` or a checkbox list in this situation.

2. There's a pixel or so worth of vertical layout shift happening in multi-select mode when there is nothing selected vs the first option is selected. This seems to be the case whether using `size` medium or small. When we're in multi-select mode, can we set a different min-height equal to the height of the tags to accommodate?

Image

Image

packages/react-components/src/components/Select/Select.css
mkernohanbc: fixed in [7bd12ca](https://github.com/bcgov/design-system/pull/643/commits/7bd12ca33937c5fc290d400edbc3f4f9b47155f4)

packages/react-components/src/components/Select/Select.css
mkernohanbc: fixed in [8301d56](https://github.com/bcgov/design-system/pull/643/commits/8301d56429b37005df7d08eef8c2844eea26806c)

packages/react-components/src/components/Select/Select.css
```diff
@@ -1,55 +1,106 @@
...
 }
 
-/* Error icon */
-.bcds-react-aria-Select[data-invalid]
-  > .bcds-react-aria-Select--Button
-  > svg:not(:last-child) {
-  color: var(--icons-color-danger);
+/* Sizing */
+.bcds-react-aria-Select.small .bcds-react-aria-Select--Button {
+  min-height: 32px;
```
ty2k, line 59: Following up on the comments [here](https://github.com/bcgov/design-system/pull/643#pullrequestreview-4047687670), I think this value needs to change to accommodate the extra ~1px happening when we're in multi-select mode with one option selected.

What I suggest is adding a class on the `Select` to indicate that you're in multi-select mode, and then choosing one of these:

1. To keep all of the fields the same height, you could start playing with the padding inside of the Select button to accommodate the extra ~1px of height the Tag causes.

2. If you don't care if the overall height of the Select field in multi-select mode is slightly taller than other components, you can set this `min-height` to 40px and the equivalent `medium` `min-height` to 46px.

ty2k (changes requested):
One visual height issue remains on this when we're in multi-select and moving between having no options selected and having one option selected.

packages/react-components/src/components/Select/Select.css
mkernohanbc: [a939764](https://github.com/bcgov/design-system/pull/643/commits/a939764238f75c340c7a5ae14c0359eda23c90ba) fixes this vertical drift in as minimally-invasive a way as I can figure. It trims 0.5px of padding from the top and bottom of the tag overlay. It's a bit of a kludge — what do you think?

ty2k: Sick, that works!

https://github.com/user-attachments/assets/55ff5871-2fa4-4926-bbcc-9cc65f2eb6c7

mkernohanbc:
@Philip-Cheung @ty2k this is ready for final review now that I've resolved various outstanding issues:

1. When the Select is in multi-select mode _and_ scrollable (fixed height), we no longer suppress the browser-default horizontal scrollbar — this was a bad call from usability/accessibility perspective
2. For now, however, I propose we don't add a visual effect on the right edge to indicate overflow/hidden items: it's feasible, but hard to do robustly and will add a bunch of additional complexity (programmatically identifying the last currently-visible item reliably, including on resize, etc.)
3. I've fixed the annoying 1px shift after an item is selected and the TagGroup renders

ty2k (changes requested):
I think this is essentially ready to go!

The change suggestions from Copilot above are all good. We don't really do any special functionality to unify `className` with user-supplied `prop.className` in this library so far. It might be good to come up with a common way of handling this - something beyond the scope of this PR, IMO.

Minor focus styling change requests below, and I'm not sure if these changes are best addressed here or https://github.com/bcgov/design-system/pull/593.

1. The border color of the `<button>` element turns blue and we're also doing the blue offset focus ring. Compare the Button component to this implementation as Select for an example.

Select:
Image

Button:
Image

2. In multi-select mode, the styling on the "x" button is different in Firefox vs Chrome:

Firefox:
Image

Chrome:
Image

mkernohanbc:
> I think this is essentially ready to go!
> 
> The change suggestions from Copilot above are all good. We don't really do any special functionality to unify `className` with user-supplied `prop.className` in this library so far. It might be good to come up with a common way of handling this - something beyond the scope of this PR, IMO.

Agreed! I'll come back to that in a later change — need to think about if/how we want to tackle allowing users to add their own classes on top of ours.
 
> 1. The border color of the `<button>` element turns blue and we're also doing the blue offset focus ring. Compare the Button component to this implementation as Select for an example.

This is consistent with how we handle focus on other input elements (see for example TextField/NumberField.) Agree there might be a bit more harmonization work to be done, but I'd like to take that out of the scope of this PR and circle back on it later.

> 2. In multi-select mode, the styling on the "x" button is different in Firefox vs Chrome:

This is because (TIL) Tag re-imports RA Button rather than using the BCDS button, so it defaults to browser focus behaviour. [218fd54](https://github.com/bcgov/design-system/pull/643/commits/218fd54cf49855c1c1ad22d7c4f90742d820f3c4) adds explicit focus behaviour derived from our button.

packages/react-components/src/components/Select/Select.tsx
```diff
@@ -59,44 +72,167 @@ export interface SelectProps<T extends object> extends ReactAriaSelectProps<T> {
...
+  const [tagOverlay, setTagOverlay] = useState<HTMLDivElement | null>(null);
+
   return (
-    <ReactAriaSelect {...props} className="bcds-react-aria-Select">
-      {({ isOpen, isRequired, isInvalid }) => (
+    <ReactAriaSelect
+      selectionMode={selectionMode}
+      className={`bcds-react-aria-Select ${size} ${overflow}`}
+      {...props}
+    >
```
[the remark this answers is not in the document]
mkernohanbc: To address systemically in future PR

packages/react-components/src/components/Tag/Tag.tsx
```diff
@@ -42,17 +42,19 @@ export default function Tag({
...
   textValue,
+  ...props
 }: TagProps) {
   return (
     <ReactAriaTag
       className={`bcds-react-aria-Tag ${color} ${tagStyle} ${size}`}
       id={id}
       textValue={textValue}
+      {...props}
     >
```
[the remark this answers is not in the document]
mkernohanbc: To address systemically in future PR

ty2k (approved):
Great work! 🥇

PR #647, opened 2026-03-19
Add Menu and Navbar components

### LINKED ISSUE #518: Menu/menu item
The menu component enables the creation of dropdown menus containing static or dynamic lists of items. It looks and behaves similarly to Select, but is intended for use in navigation and configuration contexts, rather than for data input.

A common anticipated use case is within the [Header](https://www2.gov.bc.ca/gov/content/digital/design-system/components/header) and [Footer](https://www2.gov.bc.ca/gov/content/digital/design-system/components/footer) components. 

It should support:

- Sections and headings
- Nested submenus
- Hover/focus/disabled states
- High- and low-profile styles, for different usages (ie. variant used in a Header/primary nav context should be more prominent than in a Footer/secondary navigation context)

This component will be implemented using [RAC Menu](https://react-spectrum.adobe.com/react-aria/Menu.html) and the MenuItem subcomponent.

mkernohanbc:
Draft PR for initial feedback on design and implementation.

This PR adds two new components:

- [Menu](https://react-aria.adobe.com/Menu) and various subcomponents: a dropdown menu for use in cases where a Select is inappropriate (eg. navigation)
- Navbar: a generic navigation bar, implemented based on [RA Toolbar](https://react-aria.adobe.com/Toolbar)

## Menu
Screenshot 2026-03-19 at 2 28 06 PM

Menu is logically very similar to Select, and follows the same [Collections API](https://react-aria.adobe.com/collections). 

The `MenuTrigger`, `Menu` and `MenuItem` components are the core parts, but `MenuSection`, `MenuSection` and `SubmenuTrigger` components are also exported to support complex composition:

```javascript
function App() {
  return (
    <MenuTrigger>
      <Button>This button activates the menu</Button>
      <Menu>
        <MenuSection>
          <MenuSectionHeader />
          <MenuItem />
        </MenuSection>
        <Separator />
        <MenuSection>
          <MenuSectionHeader />
          <MenuItem />
        </MenuSection>
      </Menu>
    </MenuTrigger>
  );
}
```

Menus can be composed dynamically using the `items` or `sections` props, or manually composed via the `children` slot. 

See Storybook docs for a more detailed explanation of the structure of this set of components, and each component's individual API.

## Navbar
Screenshot 2026-03-19 at 2 33 57 PM

This is a (currently extremely basic) implementation of an idea we've discussed previously: a secondary navigation menu that can be used alongside the Header component. 

What I currently have on this branch is:

- A standalone Navbar component, implemented using [RA Toolbar](https://react-aria.adobe.com/Toolbar)
- A change to Header to natively integrate Navbar via a new `subMenuItems` prop

Navbar has relatively little internal logic of its own, beyond sizing/spacing and automatically generating `<Separator />` elements between items. It is built around a generic `children` slot, expecting some number of `<Link>`, `<Button>`, `<Menu>`  etc.


packages/react-components/src/components/Navbar/Navbar.tsx
```diff
@@ -0,0 +1,46 @@
...
+              key={`sep-${index}`}
+            />,
+          ]
+        : []),
+    ]);
+  }
+
+  return (
+    <div className="bcds-react-aria-Navbar">
+      <Toolbar
```
ty2k, line 38: I think [Toolbar](https://react-aria.adobe.com/Toolbar) should just be a regular [`<nav>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/nav). 

The issue with Toolbar is it gets the [toolbar role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/toolbar_role), whereas a `<nav>` gets [navigation](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/navigation_role).

packages/react-components/src/components/Popover/Popover.tsx
```diff
@@ -0,0 +1,12 @@
+import { Popover as ReactAriaPopover } from "react-aria-components";
+import type { PopoverProps } from "react-aria-components";
+
+import "./Popover.css";
+
+export default function Popover(props: PopoverProps) {
```
ty2k, line 6: Is this a general purpose Popover or specific to Navbar and Menu? Can it go in by itself if it's generic? If it's generic, we're already using the Popover from RAC throughout the library, so the names and usage should be reconciled.

If it's specific, can it be renamed for specificity?

packages/react-components/src/components/Header/Header.tsx
```diff
@@ -1,5 +1,5 @@
 import { cloneElement, PropsWithChildren } from "react";
-
+import Navbar from "../Navbar";
```
ty2k, line 2: Can we leave it out of `Header`?

- Better separation of concerns.
- Leaves the `Header` props cleaner.
- Better customization for breakpoints (contrived example: Header goes from desktop to tablet design at 900px but Navbar needs to change at 750px).

packages/react-components/src/components/Icons/SvgCheckIcon/SvgCheckIcon.tsx
```diff
@@ -1,16 +1,16 @@
...
-export default function SvgCheckIcon({ id = "check-icon" }) {
+/* This component implements the Check icon from Font Awesome: https://fontawesome.com/v6/icons/check */
+export default function SvgCheckIcon({ ...props }) {
   return (
     <svg
-      id={id}
       width="10"
       height="10"
-      viewBox="0 0 10 10"
+      viewBox="0 0 448 512"
```
ty2k, line 7: I see the move to the v6 version of this icon has a different (not square) aspect ratio. Will this cause visual regressions in other components like `Checkbox`?

Something to think about in general for icons if we're going to mix and match versions of Font Awesome. I think the majority we have right now are following the old square ratio.

ty2k (changes requested):
Very slick! I like the Menu implementation and it seems to work really nicely with keyboard controls, including for complex cases like the Submenu.

I haven't gone through this with a fine-toothed comb because of the amount of code here, but this is a first pass. In general, I think it would be ideal to break this into several PRs:

- One for each SVG icon
- Popover by itself if it's generic
- Menu and MenuItem
- Navbar

mkernohanbc:
Closing this PR, superseded by narrower PRs #675 and #676.

PR #656, opened 2026-04-17
Map React props to Figma properties

### LINKED ISSUE #430: Implement Figma Code Connect
Figma's [Code Connect](https://help.figma.com/hc/en-us/articles/23920389749655-Code-Connect) extension provides a method to link UI components in Figma and in code. This provides significantly better integration between our design and code libraries, and should make it easier for developers working with Figma design documents to make effective use of the design system.

There is some overhead involved in configuring this integration and mapping each component in Figma to its reference implementation in the codebase. Needs further investigation.

Bookmarking for investigation: does it make sense to add `npx figma connect publish` as a step in our publish workflow?

mkernohanbc:
This PR starts the process of implementing Figma's [Code Connect API](https://help.figma.com/hc/en-us/articles/23920389749655-Code-Connect) (see #430). It adds `*.figma.tsx` files containing mappings between React props and Figma props alongside each component. It does not touch component files themselves.

Initial mappings were auto-generated using Figma's migration script, and then manually checked. This process has surfaced various inconsistencies between component and implementation in Figma and React (for example, inverted booleans and type mismatches.) These will be addressed in subsequent PRs.

Figma does not provide a method to test Code Connect implementation on a branch, so the next steps will be to:

1. Publish initial mappings to Figma
2. Review mappings in Code Connect
3. Update properties in Figma
4. Follow-up PRs to adjust mapping files


packages/react-components/figma.config.json
```diff
@@ -0,0 +1,7 @@
+{
+  "codeConnect": {
+    "include": ["**/*.figma.ts", "src/components/**/*.tsx"],
```
ty2k: I see `**/*.figma.ts` in the `include` globs here, but no new files that would match this pattern. I think this is auto-generated, so just pointing this out.

packages/react-components/package.json
```diff
@@ -51,6 +51,7 @@
     "react-dom": "^16.14.0 || ^17.0.0-rc.1 || ^18.0.0 || ^19.0.0"
   },
   "devDependencies": {
+    "@figma/code-connect": "^1.4.3",
```
ty2k: Can we please pin this to just `"1.4.3"` to match the pinning for the rest of the `devDependencies`?

ty2k (approved):
Super cool! 🦾 🤖 

I don't have a way to test this with the full Figma integration, but I've tested that:

- All scripts still run as expected
- All tests still pass
- No linting or formatting issues
- No change to the distributed code after running `build` (ex: to make sure the code from the `*.figma.tsx` files doesn't accidentally end up in the dist bundle)

There are no blockers to merging here, but I have a couple of things to note below. Also, can we please add a line or two to the README to point out what these files are for?

packages/react-components/figma.config.json
mkernohanbc: I'm pretty sure this is a typo (should be *.figma.**tsx**) — fixed in [d7f01c1](https://github.com/bcgov/design-system/pull/656/commits/d7f01c13bf4ffee32a8373d0d75d36d8c4df12b3)

packages/react-components/package.json
mkernohanbc: done in [ee8883f](https://github.com/bcgov/design-system/pull/656/commits/ee8883f22955c298a14ebe3fef381cb9261af87a)

mkernohanbc:
@ty2k this is new: rebased onto `main` to work around merge conflicts with `package-lock.json` and suddenly we have unit tests failing that weren't before and aren't failing on main (despite the fact that this PR doesn't touch any component files?)

See https://github.com/bcgov/design-system/actions/runs/24681490358/job/72179686941?pr=656

This is going to be a fun bit of diagnosis...

packages/react-components/tsconfig.types.json
```diff
@@ -6,7 +6,7 @@
     "noEmit": false,
     "outDir": "dist/types",
     "declarationDir": "dist/types",
-    "types": ["react", "react-dom"]
+    "types": ["react", "react-dom", "@figma/code-connect/figma-types"]
```
ty2k: Just noticed that this is in `tsconfig.types.json`, which is only used for building the component library. It's probably not doing anything right now. Might be okay to just leave out entirely?

ty2k (commented):
Just noticed this one while debugging:

ty2k:
I think `package-lock.json` just needs to be regenerated from scratch for this to work.

```
rm -rf node_modules
rm package-lock.json
npm i
npm run test
# Tests should pass here.
# Commit the new package-lock.json.
```

packages/react-components/tsconfig.types.json
mkernohanbc: Checked a couple of *.figma.tsx files, no change or new errors from removing this. It was included as part of the initial setup steps, but if things work without it 🤷  https://developers.figma.com/docs/code-connect/quickstart-guide/#configure-your-project

mkernohanbc: Removed in [068c05f](https://github.com/bcgov/design-system/pull/656/commits/068c05f92c8fd669061d02a4fb03d306eddd7a35)

mkernohanbc:
This is super strange, but something in here is now introducing an [eslint error](https://github.com/bcgov/design-system/actions/runs/24688343855/job/72203458512?pr=656). 

The issue seems to be with the `zod-validation-error` package used as a dependency by both `eslint-plugin-react-hooks` and `@figma/code-connect`.

`eslint-plugin-react-hooks` package.json: 
Screenshot 2026-04-20 at 1 46 03 PM

`@figma/code-connect` package.json:
Screenshot 2026-04-20 at 1 46 47 PM

How this ends up resolving for us:
Screenshot 2026-04-20 at 1 44 47 PM

@ty2k am I understanding this correctly: `@figma/code-connect` is passively forcing `eslint-plugin-react-hooks` to use `zod-validation` v3 instead of v4, which then introduces [the issue described here](https://github.com/facebook/react/issues/35045)?

The solution described in [this comment](https://github.com/facebook/react/issues/35045#issuecomment-3697572211) does seem to fix the linting issue locally. Is that a kludge we're okay with?

ty2k:
> @ty2k am I understanding this correctly: `@figma/code-connect` is passively forcing `eslint-plugin-react-hooks` to use `zod-validation` v3 instead of v4, which then introduces [the issue described here](https://github.com/facebook/react/issues/35045)?
> 
> The solution described in [this comment](https://github.com/facebook/react/issues/35045#issuecomment-3697572211) does seem to fix the linting issue locally. Is that a kludge we're okay with?

Correct, `@figma/code-connect` [currently uses](https://github.com/figma/code-connect/blob/2404b36789a18ea547dcf7154197d2e17e339200/cli/package.json#L121) `zod-validation` at "^3.2.0", which is a match for `eslint-plugin-react-hooks` at "^3.5.0 || ^4.0.0". That issue you linked to is the exact problem we're having.

I think that `overrides` block is fine to add. If you look in your `node_modules` after running `npm i`, you'll see that `zod-validation-error` goes in twice:

1. once at at v3.4.5 in `node_modules/zod-validation-error`
2. again at v4.0.2 in `node_modules/eslint-plugin-react-hooks/node_modules/zod-validation-error`

No big deal to add the `overrides` block. It can be removed down the line when/if Zod v4 support lands in the Figma library (see [this @figma/code-connect thread](https://github.com/figma/code-connect/issues/302)).

mkernohanbc:
> No big deal to add the `overrides` block. It can be removed down the line when/if Zod v4 support lands in the Figma library (see [this @figma/code-connect thread](https://github.com/figma/code-connect/issues/302)).

Okay, great — added in [233c285](https://github.com/bcgov/design-system/pull/656/commits/233c2850a4f39adf473fa02afea2b00727e40638)

PR #657, opened 2026-04-17
Update TypeScript, ESlint and React Aria

mkernohanbc:
This PR updates:

- `typescript` from 5.9.3 -> 6.0.3
- `eslint` from 9.39.2 -> 10.2.0  
- `react-aria-components` from 1.16.0 -> 1.17.0

These changes are cherry-picked from https://github.com/bcgov/design-system/pull/655#issuecomment-4271489515.

Additional changes in this PR:

- Updates various TypeScript and ESlint plugins/sub-packages to their latest versions
- Adds two React Aria sub-packages as dev dependencies to resolve a couple of new import errors that appeared in DatePicker during testing
- Removes a soon-to-be-deprecated option from `tsconfig.json` (see https://github.com/microsoft/TypeScript/issues/62508#issuecomment-3348649259), to resolve a console warning
- Adds a missing ESlint dependency that was causing linting to fail
- Adds a `rootDir` property to `tsconfig.types.json` (see https://github.com/microsoft/TypeScript/issues/62508#issuecomment-3348659946), to resolve a build error
- Updates `tsconfig.json` and `rollup.config.js`, and adds a new declarations file to resolve new TypeScript errors (TS5096 and TS2882) that popped up on build

As far as I can tell there are now no outstanding issues:

✅ test script runs successfully
✅ lint script runs successfuly
✅ build script runs successfully
✅ Storybook builds successfully
✅ Vite app builds successfully

PR #663, opened 2026-04-20
react-aria-components v1.17.0

mkernohanbc:
Separate PR for `react-aria-components` change from #657. No other changes in this PR.


ty2k (approved):
Right on ✅ Looks like `@react-aria/datepicker` and `@react-aria/i18n` are at matching/compatible versions for this. We will likely want to move all three of these together in the future whenever we can. 👍 👍

PR #667, opened 2026-04-22
Fix typo in background-color token value for InlineAlert success variant

ty2k:
Current, missing background color:
Screen Shot 2026-04-21 at 9 44 57 PM

After fixing the typo, light green background color:
Screen Shot 2026-04-21 at 9 45 16 PM


mkernohanbc (approved):
D'oh, thanks for spotting this!

PR #678, opened 2026-04-30
Add alignment and size props to AlertBanner

mkernohanbc:
This PR makes two non-breaking changes to `AlertBanner`, responding to insights surfaced by the CMS team:

## Sizing
The `size` prop controls the vertical padding and font size on the banner, to make it more flexible for use in complex/constrained UI. `size` defaults to `medium` (current behaviour), but can also be set to `small`:

Screenshot 2026-04-30 at 10 26 34 AM

## Alignment
The `alignment` prop controls the horizontal padding and the width of the `children` container. Again, this is designed to make it more flexible. `alignment` defaults to `center` (current behaviour), but can also be set to `stretch`:

Screenshot 2026-04-30 at 10 28 18 AM

### DIFF
--- a/packages/react-components/src/components/AlertBanner/AlertBanner.css
+++ b/packages/react-components/src/components/AlertBanner/AlertBanner.css
@@ -5,7 +5,6 @@
   align-items: center;
   justify-content: space-around;
   width: 100%;
-  padding: var(--layout-padding-none) var(--layout-padding-medium);
 }
 
 .bcds-Alert-Banner--Container {
@@ -14,7 +13,6 @@
   flex-direction: row;
   align-items: center;
   gap: var(--layout-padding-medium);
-  max-width: 1100px;
 }
 
 .bcds-Alert-Banner--Icon {
@@ -31,6 +29,17 @@
   flex-grow: 1;
   flex-wrap: wrap;
   color: var(--typography-color-primary-invert);
+}
+
+/* Alignment */
+.bcds-Alert-Banner.center {
+  padding: var(--layout-padding-none) var(--layout-padding-medium);
+}
+.bcds-Alert-Banner.center > .bcds-Alert-Banner--Container {
+  max-width: 1100px;
+}
+.bcds-Alert-Banner.stretch {
+  padding: var(--layout-padding-none) var(--layout-padding-large);
 }
 
 /* Sizing */
--- a/packages/react-components/src/components/AlertBanner/AlertBanner.css
+++ b/packages/react-components/src/components/AlertBanner/AlertBanner.css
@@ -15,7 +15,6 @@
   align-items: center;
   gap: var(--layout-padding-medium);
   max-width: 1100px;
-  padding: var(--layout-padding-small) var(--layout-padding-none);
 }
 
 .bcds-Alert-Banner--Icon {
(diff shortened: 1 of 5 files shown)


Philip-Cheung (approved):

packages/react-components/src/components/AlertBanner/AlertBanner.css
```diff
@@ -31,10 +28,36 @@
...
-  font: var(--typography-regular-body);
   color: var(--typography-color-primary-invert);
 }
 
+/* Alignment */
+.bcds-Alert-Banner.center {
+  padding: var(--layout-padding-none) var(--layout-padding-medium);
+}
+.bcds-Alert-Banner.center > .bcds-Alert-Banner--Container {
+  max-width: 1100px;
```
ty2k, line 39: Out of scope for this PR, but this is probably a magic number value that needs to become a token since we're using it to space things in relation to the Header constraints. I end up using it in apps that use the BCDS for layout purposes.

packages/react-components/src/components/AlertBanner/AlertBanner.tsx
```diff
@@ -7,6 +7,10 @@ import SvgExclamationIcon from "../Icons/SvgExclamationIcon";
 import SvgInfoIcon from "../Icons/SvgInfoIcon";
 
 export interface AlertBannerProps extends React.PropsWithChildren {
+  /* Sets vertical height and font size of the banner */
+  size?: "small" | "medium";
+  /* Sets width of the banner */
+  alignment?: "center" | "stretch";
```
ty2k: The prop name "alignment" and the individual named choices don't really make sense for something that's user-facing. Right now, this describes how the flexbox implementation is working internally, not what the user expects to have happen when they pass the prop.

Instead, how about `constrained` vs `fullWidth`? Or maybe `fixed` vs `fluid`?

Whatever you pick for this, make sure you like it, because you'll end up getting requests to add it to `Header`, `Footer`, and any other similar elements that have been following that `1100px` magic number.

ty2k (changes requested):
Nice! This works but the naming of the `alignment` prop and its choice names should be changed.

packages/react-components/src/components/AlertBanner/AlertBanner.tsx
mkernohanbc: [978df17](https://github.com/bcgov/design-system/pull/678/commits/978df17882ef49f11db76b04e53acb287ae2723b) renames the prop to `layout`, with options `fixed` and `fluid`. IMO that's sufficiently descriptive?

ty2k: Nice!

ty2k (approved):
🚀

packages/react-components/src/components/AlertBanner/AlertBanner.css
mkernohanbc: Worth considering, bookmarked to revisit later (#685)

packages/react-components/src/components/AlertBanner/AlertBanner.tsx
```diff
@@ -50,7 +56,10 @@ export default function AlertBanner({
   ...props
 }: AlertBannerProps) {
   return (
-    <div className={`bcds-Alert-Banner ${variant}`} {...props}>
+    <div
+      className={`bcds-Alert-Banner ${variant} ${size} ${alignment}`}
```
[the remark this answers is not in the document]
mkernohanbc: Test suite added in [b27a830](https://github.com/bcgov/design-system/pull/678/commits/b27a83087c33b8d5469e80ccfbff6e6c1f214f70)

PR #679, opened 2026-04-30
Fix Storybook console error and update tests

### LINKED ISSUE #658: NumberField Distance Units story needs label on Select
### Component

NumberField

### Description

In the React components library, when running `npm run test-storybook`, an accessibility error appears in the output related to the NumberField's `DistanceUnits` story:

```
stderr | src/stories/NumberField.stories.tsx > Distance Units
If you do not provide a visible label, you must specify an aria-label or aria-labelledby attribute for accessibility
```

It looks like this comes from the `<Select>` instance in the example that doesn't have a label:

NumberField DistanceUnits story

Beyond fixing this story, we might want a case like this to actually fail a test run rather than just appearing in the logs for `test-storybook` as a `stderr` line. This is probably something that can be configured with Storybook alongside the `@storybook/addon-a11y` plugin.

### Steps to reproduce

Run `npm run test-storybook` or examine the console when viewing the NumberField DistanceUnits story in Storybook dev mode: 

```
http://localhost:6006/?path=/story/inputs-and-controls-numberfield--distance-units
```

### Technical information

Affects the current `main` branch of the React components library.

mkernohanbc:
This PR resolves the issues identified in #658. It adds a missing `aria-label` to one of the NumberField examples. It also adds some new Vitest config that will cause both the unit test and Storybook test suites to fail on `console.warn()`. 

Without the fix in d1ca4a95eef881b7fd01ca8acbfd9d2837203758:
 
Screenshot 2026-04-30 at 1 50 59 PM

With the fix in place, all tests pass as expected.

### DIFF
--- a/packages/react-components/.storybook/vitest.setup.ts
+++ b/packages/react-components/.storybook/vitest.setup.ts
@@ -1,7 +1,24 @@
 import * as a11yAddonAnnotations from "@storybook/addon-a11y/preview";
 import { setProjectAnnotations } from "@storybook/react-vite";
+import { afterEach, beforeEach, expect, vi } from "vitest";
 import * as projectAnnotations from "./preview";
 
 // This is an important step to apply the right configuration when testing your stories.
 // More info at: https://storybook.js.org/docs/api/portable-stories/portable-stories-vitest#setprojectannotations
 setProjectAnnotations([a11yAddonAnnotations, projectAnnotations]);
+
+let warnSpy: ReturnType<typeof vi.spyOn>;
+
+beforeEach(() => {
+  warnSpy = vi.spyOn(console, "warn").mockImplementation(() => {
+    // Silence warning output during tests; assertions below enforce failure.
+  });
+});
+
+afterEach(() => {
+  try {
+    expect(warnSpy).not.toHaveBeenCalled();
+  } finally {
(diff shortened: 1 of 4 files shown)


ty2k (changes requested):
Nice! This works as written, but let's use the suggestion in the [Copilot review](https://github.com/bcgov/design-system/pull/679#pullrequestreview-4306410262) and initialize the spy outside of the `beforeEach()` block and then do `mockRestore()` in an `afterAll()`.

I've confirmed that I can cause test failures by introducing new console warnings, like rendering a `NumberField` in a unit test with no label, or re-introducing the warning in `NumberField.stories.tsx` that's fixed in d1ca4a95eef881b7fd01ca8acbfd9d2837203758.

mkernohanbc:
> Nice! This works as written, but let's use the suggestion in the [Copilot review](https://github.com/bcgov/design-system/pull/679#pullrequestreview-4306410262) and initialize the spy outside of the `beforeEach()` block and then do `mockRestore()` in an `afterAll()`.

this is helpful, seems more robust — adopted proposed pattern in [098f010](https://github.com/bcgov/design-system/pull/679/commits/098f010652da86a00af2628a83678c9c8ab7d789).

ty2k (approved):
🚀

PR #680, opened 2026-05-20
Update dependencies

mkernohanbc:
Updates all dependencies for `@bcgov/design-system-react-components` to their latest versions. See commit notes for affected packages and versions. All updates are minor or patch changes. Test/build scripts are all running successfully.


ty2k:
@mkernohanbc One thing I notice going through the tokens package, we don't have the contents of the `dist` folder updated to match what the current code in `main` actually generates. <s>Before this PR</s>, could we please get a PR cut from `main` that runs `npm run build` and `npm run prepare-npm-package`, with the changes to files in `dist` checked in?

<s>That will let us go through those in isolation and make sure everything is working as we expect before we update these deps.</s>

ty2k:
@mkernohanbc Also, thinking about this issue a bit more, this might be a thing where we actually want to enforce a pre-commit hook that runs those build scripts and stages any changes before a commit. Let's chat about how that might work!

mkernohanbc:
> @mkernohanbc One thing I notice going through the tokens package, we don't have the contents of the `dist` folder updated to match what the current code in `main` actually generates. Before this PR, could we please get a PR cut from `main` that runs `npm run build` and `npm run prepare-npm-package`, with the changes to files in `dist` checked in?
> 
> That will let us go through those in isolation and make sure everything is working as we expect before we update these deps.

See https://github.com/bcgov/design-system/pull/683 — is this what you're asking for?

Should note, though: this PR doesn't touch the deps for `@bcgov/design-tokens` at all. I was going to update those separately (IIRC there was something in a newer version of `style-dictionary` that broke something in our pipeline, that I needed to circle back and investigate.)

ty2k (approved):
LGTM 🚀

PR #687, opened 2026-05-25
Add .npmrc for both libs

mkernohanbc:
This PR adds `.npmrc` config for both `design-tokens` and `design-system-react-components`, to set GitHub Packages as the registry of record for `@bcgov`-scoped dependencies.

This is necessary to prevent future upstream dependency issues when installing either of these packages from GitHub Packages.

**Note**: hold on merging this PR until:

- [x] `@bcgov/bc-sans` is available on GitHub Packages
- [x] New versions of `@bcgov/design-tokens` and `@bcgov/design-system-react-components` are staged for release
- [x] `@bcgov/design-system-react-components` is updated to use the latest version of `design-tokens`

### DIFF
--- a/.github/workflows/publish_design_tokens.yaml
+++ b/.github/workflows/publish_design_tokens.yaml
@@ -228,8 +228,8 @@ jobs:
           registry-url: "https://npm.pkg.github.com/"
           scope: "@bcgov"
 
-      - name: Publish to GitHub Packages @latest
-        run: npm publish --tag latest --registry=https://npm.pkg.github.com/
+      - name: Publish to GitHub Packages
+        run: npm publish --registry=https://npm.pkg.github.com/
         working-directory: ./packages/design-tokens/dist
         env:
           NODE_AUTH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/publish_design_tokens.yaml
+++ b/.github/workflows/publish_design_tokens.yaml
@@ -228,8 +228,8 @@ jobs:
           registry-url: "https://npm.pkg.github.com/"
           scope: "@bcgov"
 
-      - name: Publish to GitHub Packages
-        run: npm publish --registry=https://npm.pkg.github.com/
+      - name: Publish to GitHub Packages @latest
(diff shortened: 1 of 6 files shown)


ty2k:
Random thoughts on this one:

- The `.npmrc` is going to end up being specific to each publishing workflow, I think. Maybe it makes sense to have multiple `.npmrc` configurations depending on the workflow, and you could have a step that copies/renames the thing appropriately for each workflow, one for npm, one for GitHub Packages? Or perhaps avoid the file entirely and use the [`registry` parameter in the `actions/setup-node` action](https://github.com/actions/setup-node#usage) if that's all that needs configuring?

- `min-release-age` is a great idea for the publish workflows. In practice in local dev, I find I often want to comment it out or add exceptions when juggling new dependencies. No idea what's best here yet since this dependency cooldown idea is new in npm land.

- Unrelated but in the same vein, npm recently added [staged publishing](https://docs.npmjs.com/staged-publishing). I've tried this out with the OIDC flow and it's kind of slick. You do `npm stage publish` instead of `npm publish`, and then when you want to publish, you log into npm and authorize that one publish instance. Basically another guardrail against poisoned GitHub Actions publishing malicious versions of packages. If we wanted to adopt this in the npm publishing workflow, we would need to install npm globally at v11.15.0 or higher after the `setup-node` stage (`npm install -g npm@11.16.0`).

ty2k (approved):
This can be merged, but I think once you do, you might find you need to go back and manually publish `@bcgov/design-tokens` to GitHub Packages, at least the version needed by the React components. I think right now, if you tried to publish a new version of the components, it will fail because the tokens package is a dependency that's not available on GitHub Packages yet. Take a look at the [BC Sans GitHub Packages publishing workflow](https://github.com/bcgov/bc-sans/blob/main/.github/workflows/publish-github-package.yaml) to see how I did it there.

mkernohanbc:
Moving this back to draft for now, pending a bit more investigation after current releases are done.

mkernohanbc:
Closing this for now; unclear that it's necessary and/or what the clear benefit is at this point.

PR #688, opened 2026-05-27
Update style-dictionary and token formatting

### LINKED ISSUE #661: Update Style Dictionary
We are stuck a couple of minor versions back on `style-dictionary`, until we can diagnose and resolve a new issue that crops up in the build process when we use v5.4.0:

> I'm getting erroneous output when switching between these versions after running the build script and examining the output. Let's break the `design-tokens` changes out to another branch to diagnose.
> 
> With `style-dictionary` 5.2.0:
> 
> ```
> // variables.css
> --typography-regular-display: 400 3rem/AUTO 'BC Sans'; /** Extra-large body text size. Do not use for headings */
> ```
> 
> With the new `style-dictionary` 5.4.0:
> 
> ```
> // variables.css
>   --typography-regular-display: 400 3rem/NaNAUTO 'BC Sans'; /** Extra-large body text size. Do not use for headings */
>                                          ^^^
> ```
> 
> See the `NaN` getting prepended to `AUTO` above.
> 
> We should add a specific test for these composite font checks that use `AUTO`, similar to [this one for `--theme-blue-100`](https://github.com/bcgov/design-system/blob/main/packages/design-tokens/build-output.test.js#L53).

_Originally posted by @ty2k in https://github.com/bcgov/design-system/pull/655#discussion_r3105527875_

mkernohanbc:
This PR updates `style-dictionary` to its latest version (v5.4.4), and resolves some new issues that have cropped up in the `design-tokens` build pipeline. 

For the sake of caution, this change will necessitate a major version (v5.0) release.

## Line height
`style-dictionary` update created a new issue related to the `typography.lineHeights.auto` token, when used in composite typography tokens. 

This token previously had the value `AUTO`, [inherited from Figma](https://docs.tokens.studio/manage-tokens/token-types/typography/line-height#possible-values), which is actually a long-standing error: `line-height: auto` is not valid CSS (should be `line-height: normal`).

The issue here is that `style-dictionary` now expects the line-height value in a composite token to be a number. This doesn't cause transformation to fail, but does mean the output is deformed: 

```bcdsTypographyItalicDisplay: "400 italic 3rem/NaNAUTO 'BC Sans'",```

This token is not widely used: only the `display` tokens consume it internally, and they themselves aren't widely used (e.g. none of our components use them.) This PR makes two changes to fix this issue narrowly:

- ~2e2246e076d99366201cd776b7e9108e309a5df0 changes value of `typography.lineHeights.auto` to `normal`~ See https://github.com/bcgov/design-system/pull/688#discussion_r3359107573 
- c8f3927de967b8d31f7c4da394c9d0333d52541c removes explicit line height from `display` tokens, which will default to `normal` and side-step the `NaN` issue

We'll need to investigate how to fix this more durably upstream in Tokens Studio.

## Transform groups
e47fe1e7be58a4cb0af2f1bca9da010f15325047 removes some redundant legacy custom transforms, which triggers some change to token formatting:

- Do not force capitalization of hex codes
- Re-order composite typography tokens to `style-dictionary` default, matching the `font` shorthand

[717faa](https://github.com/bcgov/design-system/pull/688/commits/1717faa39f616e3518ddfb291369bf26918d9010) adds one new custom transform to programmatically rewrite Figma's `AUTO` value for line height to `normal`

### CODE REVIEW

File: packages/design-tokens/package.json [hunk 1]
```diff
@@ -29,6 +29,6 @@
   "homepage": "https://github.com/bcgov/design-system#readme",
   "dependencies": {
     "@tokens-studio/sd-transforms": "2.0.3",
-    "style-dictionary": "5.2.0"
+    "style-dictionary": "5.4.1"
   }
```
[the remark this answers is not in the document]
mkernohanbc: release staging will happen in a separate PR

packages/design-tokens/input/tokens.json
```diff
@@ -635,7 +635,7 @@
         "description": "Largest line height"
       },
       "auto": {
-        "value": "AUTO",
+        "value": "normal",
```
ty2k: Just wondering on the changes to `input/tokens.json` - are these manual changes to fix the `NaN` issue with composite font tokens? I'm guessing this is not a new extract from Tokens Studio because of the license issue but just want to be sure.

I find we don't actually reference `typography.lineHeights.auto` directly in any composite tokens after the rest of the changes in this file. So besides fixing the `NaN` issue, we're just outputting the string `"normal"` instead of the string `"AUTO"` for this token, right? That's the only side effect of this change?

No problem with any of this, just trying to get what's happening straight in my head.

packages/design-tokens/build-output.js
```diff
@@ -228,50 +115,38 @@ const sd = new StyleDictionary({
...
-        "bcds/css/size/zero",
-        "name/kebab",
-      ],
+      transforms: ["name/kebab"],
       buildPath: "build/scss/",
       files: [
         {
           destination: "variables.scss",
           format: "scss/map-deep",
-          filter: (t) => (
```
ty2k, line 242: Beyond the needs of this PR, but it might be useful to add the Prettier config to this project and add `prettier` as a devDependency just to keep all the formatting explicit.

packages/design-tokens/build/js/index.d.ts
```diff
@@ -136,12 +136,12 @@ export const iconsColorSuccess: string;
 export const iconsColorWarning: string;
 /** Default colour for icons on dark backgrounds */
 export const iconsColorPrimaryInvert: string;
-export const layoutBorderWidthNone: string;
+export const layoutBorderWidthNone: number;
```
ty2k: Looks like the way zero values `0` are handled is slightly different now, I think based on what happens in the `ts/size/px` transformation. It used to be that `"0"` would be outputted as a string and now it's a number. Another good reason for the major version bump, I think.

packages/design-tokens/build-output.js
```diff
@@ -3,93 +3,6 @@ import StyleDictionary from "style-dictionary";
 
 register(StyleDictionary);
 
-// TODO: Remove the `bcds` custom transformations and keep the default new
```
ty2k, line 6: Great opportunity to cut this stuff out and get back to the default path! 👏

ty2k (approved):
Great work! A few questions and comments below, but no blockers to merging. 🚀

packages/design-tokens/input/tokens.json
mkernohanbc: Yes, this is working with the last-used Tokens Studio export, until we can get back in and make changes at source.

https://github.com/bcgov/design-system/commit/c8f3927de967b8d31f7c4da394c9d0333d52541c is the change that directly fix the NaN issue (by no longer setting an explicit line-height in the affected `font` tokens.)

https://github.com/bcgov/design-system/commit/2e2246e076d99366201cd776b7e9108e309a5df0 is intended to make the `typography.lineHeights.auto` token usable on its own. Now that I'm thinking about it, though, perhaps a better approach here is a custom transform to rewrite Figma's `AUTO` to `normal` during transformation. I'm not 100% sure we'll be able to fix this elegantly in Tokens Studio.

packages/design-tokens/build-output.js
mkernohanbc: Good idea, noted in #691 to revisit later.

packages/design-tokens/input/tokens.json
mkernohanbc: @ty2k I think I have a more robust solution here now:

- [1717faa](https://github.com/bcgov/design-system/pull/688/commits/1717faa39f616e3518ddfb291369bf26918d9010) adds a custom transform to rewrite the value `AUTO` from Figma to `normal` and tests for this case
- [43a3958](https://github.com/bcgov/design-system/pull/688/commits/43a39586d37116e44c522350aa5ad6b75ca6aa2a) and [651e0bd](https://github.com/bcgov/design-system/pull/688/commits/651e0bd77853913a1834517630641f0b94b87edd) revert the manual change to the input file and previous rebuild, to get us back to previous state
- [e0f22eb](https://github.com/bcgov/design-system/pull/688/commits/e0f22eb1aad1da88725ddcfa24413b9a91aec9e8) re-runs the build to validate that the output is correct

This block in `tokens.json`: 

```
  "typography": {
    "lineHeights": {
      "auto": {
        "value": "AUTO",
        "type": "lineHeights",
        "description": "System token, used to generate type styles"
      }
```

Is now transformed to:

CSS: 
`--typography-line-heights-auto: normal; /** System token, used to generate type styles */`

JS: 
`export const typographyLineHeightsAuto = "normal"; // System token, used to generate type styles`

Thoughts?

packages/design-tokens/build/js/index.d.ts
mkernohanbc: On further investigation, this looks like a regression that was fixed in a subsequent `style-dictionary` release: https://github.com/style-dictionary/style-dictionary/releases/tag/v5.4.2

[e88d42d](https://github.com/bcgov/design-system/pull/688/commits/e88d42d5cb12dcf32ce4190a50394cadd2a6ef2a) bumps `style-dictionary` to 5.4.4 and re-runs the build. These values now revert to their previous states.

packages/design-tokens/input/tokens.json
ty2k: Awesome stuff @mkernohanbc  :rocket:

packages/design-tokens/build/js/index.d.ts
ty2k: Great find! ✅

ty2k (approved):
🚀🚀🚀

PR #690, opened 2026-05-29
Add Menu, MenuItem and MenuTrigger components

### LINKED ISSUE #518: Menu/menu item

mkernohanbc:
Forking WIP on Menu component from #675 into a clean PR.

## Menu
Screenshot 2026-03-19 at 2 28 06 PM

Menu is logically very similar to Select, and follows the same [Collections API](https://react-aria.adobe.com/collections). 

The `MenuTrigger`, `Menu` and `MenuItem` components are the core parts, but `MenuSection`, `MenuSection` and `SubmenuTrigger` components are also exported to support complex composition:

```javascript
function App() {
  return (
    <MenuTrigger>
      <Button>This button activates the menu</Button>
      <Menu>
        <MenuSection>
          <MenuSectionHeader />
          <MenuItem />
        </MenuSection>
        <Separator />
        <MenuSection>
          <MenuSectionHeader />
          <MenuItem />
        </MenuSection>
      </Menu>
    </MenuTrigger>
  );
}
```

Menus can be composed dynamically using the `items` or `sections` props, or manually composed via the `children` slot. 

See Storybook docs for a more detailed explanation of the structure of this set of components, and each component's individual API.

### DIFF
--- a/packages/react-components/src/components/Menu/Menu.css
+++ b/packages/react-components/src/components/Menu/Menu.css
@@ -43,3 +43,4 @@
   padding: var(--layout-padding-xsmall) var(--layout-padding-xsmall);
   min-width: var(--trigger-width);
   overflow-y: auto;
+}
--- a/packages/react-components/src/components/Menu/Menu.css
+++ b/packages/react-components/src/components/Menu/Menu.css
@@ -1,7 +1,8 @@
 .bcds-react-aria-Menu {
-  display: flex;
-  flex-direction: column;
-  padding: var(--layout-padding-none);
+  overflow-y: auto;
+  max-height: inherit;
+  scroll-padding-block: var(--layout-border-width-medium);
+  padding: var(--layout-padding-hair) var(--layout-padding-none);
 }
 
 /* Sizing */
@@ -28,9 +29,6 @@
 .bcds-react-aria-Menu[data-focused] {
   outline: none;
 }
-.bcds-react-aria-Popover:focus {
-  outline: none;
-}
 
 /* Popover */
 .bcds-react-aria-Popover {
@@ -40,7 +38,9 @@
   border-radius: var(--layout-border-radius-medium);
   box-shadow: var(--surface-shadow-medium);
   box-sizing: border-box;
-  padding: var(--layout-padding-xsmall) var(--layout-padding-xsmall);
-  min-width: var(--trigger-width);
-  overflow-y: auto;
+  overflow: hidden;
+  padding: var(--layout-padding-hair) var(--layout-padding-xsmall);
+  width: var(
+    --trigger-width
+  ); /* Variable provided by Select component https://react-spectrum.adobe.com/react-aria/Select.html#popover-1 */
 }
--- a/packages/react-components/src/components/Menu/Menu.css
+++ b/packages/react-components/src/components/Menu/Menu.css
@@ -1,0 +1,46 @@
+.bcds-react-aria-Menu {
+  display: flex;
+  flex-direction: column;
+  padding: var(--layout-padding-none);
+}
+
+/* Sizing */
+.bcds-react-aria-Menu.small {
+  font: var(--typography-regular-small-body);
+}
+
+.bcds-react-aria-Menu.medium {
+  font: var(--typography-regular-body);
+}
+
+/* Sections */
+.bcds-react-aria-Menu .react-aria-Header {
+  padding: var(--layout-padding-small);
+  font: var(--typography-bold-small-body);
+}
+.bcds-react-aria-Menu .react-aria-MenuSection {
+  display: flex;
+  flex-direction: column;
+  gap: var(--layout-margin-small);
+}
+
+/* States */
+.bcds-react-aria-Menu[data-focused] {
+  outline: none;
+}
+.bcds-react-aria-Popover:focus {
+  outline: none;
+}
+
+/* Popover */
+.bcds-react-aria-Popover {
+  background-color: var(--surface-color-forms-default);
+  border: var(--layout-border-width-small) solid
+    var(--surface-color-border-default);
+  border-radius: var(--layout-border-radius-medium);
+  box-shadow: var(--surface-shadow-medium);
+  box-sizing: border-box;
+  padding: var(--layout-padding-xsmall) var(--layout-padding-xsmall);
+  min-width: var(--trigger-width);
+  overflow-y: scroll;
+}
--- a/packages/react-components/src/components/Menu/Menu.css
+++ b/packages/react-components/src/components/Menu/Menu.css
@@ -1,8 +1,8 @@
 .bcds-react-aria-Menu {
+  box-sizing: border-box;
   overflow-y: auto;
   max-height: inherit;
-  scroll-padding-block: var(--layout-border-width-medium);
-  padding: var(--layout-padding-hair) var(--layout-padding-none);
+  padding: var(--layout-padding-xsmall);
 }
 
 /* Sizing */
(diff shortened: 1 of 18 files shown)


mkernohanbc:
Final outstanding issue flagged by @ty2k in https://github.com/bcgov/design-system/pull/675#discussion_r3295897058: 

> In general, I find scrolling the menu doesn't work with the keyboard very well. The scroll window "gets stuck" and it's possible to focus on elements that are totally out of view.
> 
> Example without sections:
> 
>  Screen.Recording.2026-05-24.at.9.10.33.PM.mov 
> 
> Example with sections:
> 
>  Screen.Recording.2026-05-24.at.9.11.05.PM.mov 
> 
> Contrast that behavior with how Select works when scrolling with the keyboard. Maybe the fact that Menu selection "rolls over" so you can go from the last menu item back to the first causes some weirdness?
> 
> Check out the [RAC Menu docs Section example](https://react-aria.adobe.com/Menu#sections), which seems to work fine for me. I can't spot the difference in what they're doing there that we're not, or vice versa.
> 
> We might be able to handle this with something like a `useRef()` ref combined with [`scrollIntoView()`](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView), but ideally we wouldn't have to do something like that. There might be other examples out there of folks solving this on the Spectrum repo.

packages/react-components/src/stories/Menu.mdx
```diff
@@ -0,0 +1,208 @@
...
+
+<Source
+  code={`import { Menu, MenuTrigger } from "@bcgov/design-system-react-components";`}
+  language="typescript"
+/>
+
+Learn more about working with the Menu component:
+
+- [Usage and best practice guidance](#)
+- [View the Menu component in Figma](https://www2.gov.bc.ca/gov/content?id=8E36BE1D10E04A17B0CD4D913FA7AC43#designers)
```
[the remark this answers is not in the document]
mkernohanbc: added in [14b891b](https://github.com/bcgov/design-system/pull/690/commits/14b891b29fc362cb2180a0a08ab3afefb910d2db)

mkernohanbc:
> Final outstanding issue flagged by @ty2k in [#675 (comment)](https://github.com/bcgov/design-system/pull/675#discussion_r3295897058):
> 
> > In general, I find scrolling the menu doesn't work with the keyboard very well. The scroll window "gets stuck" and it's possible to focus on elements that are totally out of view.

I think I have a fix for this in [68b6882](https://github.com/bcgov/design-system/pull/690/commits/68b6882ea740ec961896721c52b9b99c80c737cd):

- Explicitly mark the `<ReactAriaMenu>` element as the scrollable area via [scrollRef](https://react-aria.adobe.com/Popover#popover) prop
- Tweak some styling to remove some redundancies and make its styling more consistent with the approach we've already taken in Select (eg. removing focus ring on individual listbox items in favour of using the hover effect for focus)

Taken together, this seems to restore proper keyboard navigation functionality, including scrolling:

https://github.com/user-attachments/assets/f122623a-3d01-4867-a576-d488ddbcfff7

packages/react-components/src/components/Menu/Menu.tsx
```diff
@@ -0,0 +1,113 @@
...
+  header?: string;
+  /* Array of items in the section */
+  items: MenuItemProps[];
+}
+
+export interface MenuProps<
+  T extends MenuItemProps,
+> extends ReactAriaMenuProps<T> {
+  /* Set size of menu items (does not affect the MenuTrigger element) */
+  size?: "small" | "medium";
```
ty2k: Even though the Storybook docs and the comment here spell it out, my brain wants `size` to apply to the trigger button, not the items. Different name maybe? `itemSize` or something? Your call!

packages/react-components/src/components/Menu/Menu.css
```diff
@@ -0,0 +1,46 @@
...
+  background-color: var(--surface-color-forms-default);
+  border: var(--layout-border-width-small) solid
+    var(--surface-color-border-default);
+  border-radius: var(--layout-border-radius-medium);
+  box-shadow: var(--surface-shadow-medium);
+  box-sizing: border-box;
+  overflow: hidden;
+  padding: var(--layout-padding-hair) var(--layout-padding-xsmall);
+  width: var(
+    --trigger-width
```
ty2k, line 43: IMO this needs to be a `min-width` rule rather than `width`. Most menu buttons are going to be narrow, often only the hamburger icon. Check out the "Items" story example:

With `width`: 
Image

Same value but with `min-width`:
Image

Outside of this PR, it's probably worth having a think about this in Select too. It's less noticeable there because the placeholder text usually create a wider trigger width to work from, and often designs will call for the Select to be stretched to the available width, unlike a menu button.

Select with `width` rule:
Image

Select with `min-width` rule: (might even want a `max-width` rule too)
Image

packages/react-components/src/components/Menu/Menu.css
```diff
@@ -0,0 +1,46 @@
+.bcds-react-aria-Menu {
+  overflow-y: auto;
```
ty2k, line 3: Instead of scrolling the Menu, can we scroll the Popover instead?

Since we're applying padding inside of the Popover, having the Menu be the thing that scrolls pushes the scrollbar inside the padding. It also makes the `max-height: inherit` rule below problematic, because you're telling Menu to take the height of Popover, but pushing it with padding, so its bottom edge is clipped:

Image

Contrast this with Select:
Image

ty2k (changes requested):
Getting very close!

The improved keyboard navigation works good but might need a slight re-write if moving the scrolling behavior to the Popover from the Menu.

Beyond this PR, I wonder if we can get that same style of keyboard navigation going in the Select component? I find that at this point, Select has the better scroll styling (scrolling the Popover, not the Menu inside of it), but Menu has the better behavior in terms of keeping the focused element in view.

packages/react-components/src/components/Menu/Menu.tsx
mkernohanbc: good call — [7e62dd2](https://github.com/bcgov/design-system/pull/690/commits/7e62dd254e5ccc88f3e24d89c7c8dd8aace0c03d) renames this prop to `itemSize`

packages/react-components/src/components/Menu/Menu.css
mkernohanbc: added in [e2fe7b8](https://github.com/bcgov/design-system/pull/690/commits/e2fe7b81d3ea4a597a7a37bfde5bc7b0161b6747) — I've refrained from setting `max-width` for now, as I'm not sure what a sane setting looks like there.

I'll circle back on rationalization with Select after #643 is merged.

packages/react-components/src/components/Menu/Menu.css
mkernohanbc: Thanks for helping debug this. [7771713](https://github.com/bcgov/design-system/pull/690/commits/7771713b6c4ccb929075907cc87fdb7fa799e37e) resolves this (I think) satisfactorily and simply, so that keyboard nav and scroll behaviour are working as expected.

ty2k (approved):
Awesome work @mkernohanbc 👏

PR #692, opened 2026-06-02
Remove redundant tag from GitHub Packages publish job

mkernohanbc:
This change removes the dist-tag (`--tag latest`) from the GitHub Packages publish jobs for new production versions. Unlike npm, this `latest` tag doesn't have a specialized role on GitHub Packages, and including it may cause issues. My suspicion is that:

- `npm install @bcgov/design-system-react-components@latest` would work
- `npm install @bcgov/design-system-react-components` without additional scope would not work as expected


ty2k (approved):
:+1: :+1:

PR #694, opened 2026-06-11
Add Dependabot config

mkernohanbc:
This PR adds new [Dependabot](https://github.com/dependabot) config to give us a more structured approach to managing updates to the dependencies for `design-system-react-components` and `design-tokens`.

It configures Dependabot to perform a weekly scan for new versions of dependencies.

It applies two conditions:

- Only open PRs for minor/patch versions, ignore major versions (we handle these manually)
- Wait 7 days after a new version has released before opening a PR (as a defense against supply-chain attacks)


ty2k (approved):
Right on! :+1: :+1:

PR #759, opened 2026-06-25
Menu: fix TypeScript error

mkernohanbc:
This PR fixes a TypeScript error in the Menu component. Previously, there was an inconsistency in how the `items` prop (used to pass an array of menu items) was typed in the `MenuSectionProps` and `MenuProps` interfaces. In `MenuProps` it was typed as `items?: T[]`, which caused an incompatibility error with the `onAction` callback. 

With this change, `items` is now typed consistently as `items: MenuItemProps[]` in both contexts. This resolves the error, and is consistent with how we've approached this same functionality in Select.

### DIFF
--- a/packages/react-components/src/components/Menu/Menu.tsx
+++ b/packages/react-components/src/components/Menu/Menu.tsx
@@ -22,7 +22,7 @@ export interface MenuSectionProps {
   /* Text label for the section */
   header?: string;
   /* Array of items in the section */
-  items: MenuItemProps[];
+  items?: MenuItemProps[];
 }
 
 export interface MenuProps extends ReactAriaMenuProps<MenuItemProps> {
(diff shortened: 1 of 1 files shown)


ty2k (changes requested):
Good change, I think it's better to avoid the generic `<T>`.

On the `items`prop, I see right now it's required for `MenuSectionProps` but optional for `MenuProps`. If that's by design, can we add a comment to explain why? If it's not, can we use the same behaviour in both places?

mkernohanbc:
> Good change, I think it's better to avoid the generic `<T>`.
> 
> On the `items`prop, I see right now it's required for `MenuSectionProps` but optional for `MenuProps`. If that's by design, can we add a comment to explain why? If it's not, can we use the same behaviour in both places?

Good spot — this was matching the structure from Select (where `items` is the only way to compose within `sections`), but doesn't fit the design here (where we're also exporting `MenuSection` and `MenuItem` to support another way to compose.) 

[6906096](https://github.com/bcgov/design-system/pull/759/commits/690609657e37c106296136ca054d41db8d2e2f32) makes `items` optional.

ty2k (approved):
:+1: :+1:

PR #760, opened 2026-06-26
Refactor icon components

### LINKED ISSUE #436: Add accessible titles and aria-hidden support to React icon components
For commonly-used SVG icons that we implement (or have already implemented) as React components, we should explore adding support for:

- [the `<title>` attribute](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/title)
- `aria-hidden` and/or `aria-labelledby`

Without these, I suspect we risk allowing accessibility violations to creep in as we reuse these icon components.

In terms of basic implementation, I think this looks something like:

- add `title` prop to populate `<title>` attribute on the `<svg>` element
- fallback to `aria-hidden="true"` if no title is provided

@ty2k I'm happy to take a first pass at implementing this myself, but would appreciate your thoughts on whether my assumptions here are valid?

### LINKED ISSUE #614: Icon components: standardize props and styling
We need to do some audit and cleanup on our various icon components to ensure that:

- They have a consistent set of props (eg. for setting `id`)
- They are styled as efficiently as possible
- They are accessible (eg. via applying `aria-hidden` where appropriate)

For reference:

> We probably don't want to be following this pattern from the Header and Footer logo SVGs of assigning a default `id` to SVGs unless we are very sure that there should only be one instance of them that will occur on a single page. `id` is meant to be globally unique on an entire HTML page, so we should generally opt for styling SVGs using `class` where we have the option. Keeping the `id` prop as an option is nice for icons we anticipate folks importing from the library and using (like from an SVG icon/logo collection we expose proactively), though we could just [use spread `...props`](https://react.dev/learn/passing-props-to-a-component#forwarding-props-with-the-jsx-spread-syntax) for this purpose instead, IMO.

_Originally posted by @ty2k in https://github.com/bcgov/design-system/pull/611#discussion_r2691689646_

mkernohanbc:
This PR closes #436 and #614. It refactors our entire set of icon components to:

1. Add props for each component to set `size` and `title` and leverage `React.SVGProps`
2. Support basic accessibility requirements via `role`, `aria-labelledby`, `aria-hidden` and `focusable` attributes

## Rationale
These icon components are primarily intended to be consumed by other components, which own the accessibility context. However, we are also exporting them as standalone components, meaning people may use them elsewhere — this change makes them more robust and less likely to cause accessibility issues when doing so.

## Sizing
Sizing for most icon components is now handled via a new `getIconSize` utility. This maps the `size` prop on each component to the correct design token value (`iconsSizeXsmall` to `iconsSizeXlarge`.) 

These components are currently excluded from the new sizing pattern, because they have non-uniform aspect ratios and/or specialized use-cases:

- `SvgBcLogo`
- `SvgDashIcon`
- `SvgTooltipArrowUp`

## Titles and roles
Icon components now support a `title` prop, which populates a [<title>](https://developer.mozilla.org/en-US/docs/Web/SVG/Reference/Element/title) element. `<title>` receives an ID generated via [useId](https://react.dev/reference/react/useId), which is used to set `aria-labelledby` (there are some wrinkles here — see https://github.com/bcgov/design-system/pull/760#discussion_r3484541724.)

The presence or absence of `title` is a trigger for various other attributes:

- If `title` is present, the icon is considered non-decorative. `role="img"` and `aria-labelledby` are set automatically
- If title is absent, `role` is not set and `aria-hidden` is applied


packages/react-components/src/components/Icons/SvgChevronRightIcon/SvgChevronRightIcon.tsx
```diff
@@ -1,14 +1,37 @@
...
+  size?: IconSize;
+  /** Accessible label for non-decorative usage. When omitted, the icon is hidden from assistive technology. */
+  title?: string;
+}
+
+export default function SvgChevronRightIcon({
+  size = "medium",
+  title,
+  id,
+  ...props
```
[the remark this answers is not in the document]
mkernohanbc: Fixed in [40ef335](https://github.com/bcgov/design-system/pull/760/commits/40ef3357ad4ee77319db1bb7859be3f430460561); reverted to using `width` and `height` presentation attributes instead of inline `style`.

packages/react-components/src/components/Icons/SvgPlusIcon/SvgPlusIcon.tsx
```diff
@@ -1,13 +1,36 @@
...
+  size?: IconSize;
+  /** Accessible label for non-decorative usage. When omitted, the icon is hidden from assistive technology. */
+  title?: string;
+}
+
+export default function SvgPlusIcon({
+  size = "medium",
+  title,
+  id,
+  ...props
```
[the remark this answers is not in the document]
mkernohanbc: Fixed by [40ef335](https://github.com/bcgov/design-system/pull/760/commits/40ef3357ad4ee77319db1bb7859be3f430460561)

packages/react-components/src/components/Icons/SvgCheckIcon/SvgCheckIcon.tsx
```diff
@@ -1,13 +1,36 @@
...
+  size?: IconSize;
+  /** Accessible label for non-decorative usage. When omitted, the icon is hidden from assistive technology. */
+  title?: string;
+}
+
+export default function SvgCheckIcon({
+  size = "medium",
+  title,
+  id,
+  ...props
```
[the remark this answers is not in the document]
mkernohanbc: Fixed in [1a39586](https://github.com/bcgov/design-system/pull/760/commits/1a39586fccc3150bc21079cfbeee12403718b497)

packages/react-components/src/components/Icons/SvgUpRightFromSquareIcon/SvgUpRightFromSquareIcon.tsx
```diff
@@ -1,22 +1,36 @@
 /* The component implements the Up Right From Square icon from Font Awesome: https://fontawesome.com/icons/up-right-from-square */
 
-export interface SvgUpRightFromSquareIconProps {
-  size?: 16 | 20;
-  id?: string;
+import { useId } from "react";
+import getIconSize, { IconSize } from "../../../hooks/getIconSize";
```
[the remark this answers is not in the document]
mkernohanbc: @ty2k following on from our conversation yesterday, [1dbd467](https://github.com/bcgov/design-system/pull/760/commits/1dbd467aa17dec7d0fdec45893b3d29aa0e58a41) replaces React's native useId with [React Aria's useId implementation](https://react-aria.adobe.com/useId). There is one dimension to this change that gives me pause: in a React v16 or v17 project using SSR, this approach requires the use of RA's [SSRProvider](https://react-aria.adobe.com/SSRProvider). Curious what your thoughts on the tradeoffs here are?

packages/react-components/src/hooks/getIconSize.ts
```diff
@@ -0,0 +1,13 @@
+export type IconSize = "xsmall" | "small" | "medium" | "large" | "xlarge";
+
+const iconSizeMap: Record<IconSize, number> = {
+  xsmall: 14, // var(--icons-size-xsmall)
```
ty2k: Instead of hard-coding these values, let's reference them from the JavaScript version of the tokens:

```ts
import * as tokens from "@bcgov/design-tokens/js"

// ...

const iconSizeMap: Record<IconSize, string> = {
  xsmall: tokens.iconsSizeXsmall, // var(--icons-size-xsmall)
  // ...
```

You'll have to change the type signature of the utility function since you'll be returning strings like `"16px"` instead of a number `16`.

Browsers should interpret these values the same way in the context of an `<svg>` `height` and `width`.

packages/react-components/src/hooks/getIconSize.ts
```diff
@@ -0,0 +1,13 @@
...
+  small: 16, // var(--icons-size-small)
+  medium: 20, // var(--icons-size-medium)
+  large: 24, // var(--icons-size-large)
+  xlarge: 32, // var(--icons-size-xlarge)
+};
+
+export default function getIconSize(size: IconSize = "medium"): number {
```
ty2k: Since this is a utility function and not technically a hook, can we please move it into a new `src/utils` or similar directory? We should leave this `hooks` directory for genuine hooks (functions that start with `use` that follow [the rules of hooks](https://react.dev/reference/rules/rules-of-hooks) and use React hooks like `useState()` interally).

packages/react-components/src/components/Icons/SvgUpRightFromSquareIcon/SvgUpRightFromSquareIcon.tsx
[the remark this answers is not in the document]
ty2k: Good flag, I need to have a think on this one.

packages/react-components/rollup.config.js
```diff
@@ -28,7 +28,7 @@ export default [
       commonjs(),
       typescript({
         tsconfig: "./tsconfig.build.json",
-        include: ["src/index.ts", "src/components/**/*"],
+        include: ["src/index.ts", "src/components/**/*", "src/hooks/**/*"],
```
ty2k: Might need updating again after moving `getIconSize()` into a `utils` directory (or whatever name).

ty2k (changes requested):
Looking good, a few change requests below. Let me have a think on the `useId()` issue with `<SSRProvider>`. That might be an okay thing, but I need to test a bit and think of the ramifications.

packages/react-components/src/hooks/getIconSize.ts
mkernohanbc: Done in [f0fb3cc](https://github.com/bcgov/design-system/pull/760/commits/f0fb3ccb3eb7d8900ff880e8ab9838e50e911bda)

packages/react-components/rollup.config.js
mkernohanbc: Done in [d547774](https://github.com/bcgov/design-system/pull/760/commits/d5477746be09d4dc810e2fa57c636b8331dfaf71)

packages/react-components/src/hooks/getIconSize.ts
mkernohanbc: Good shout, no idea why I didn't think of this. This is done in [24f8ec8](https://github.com/bcgov/design-system/pull/760/commits/24f8ec8a99b48f414d091ca99d7137038e19c5b0). 

Browser does indeed interpret the `##px` strings and size icons correctly:
 
Screenshot 2026-07-10 at 4 31 39 PM

ty2k:
> @ty2k following on from our conversation yesterday, [1dbd467](https://github.com/bcgov/design-system/pull/760/commits/1dbd467aa17dec7d0fdec45893b3d29aa0e58a41) replaces React's native useId with [React Aria's useId implementation](https://react-aria.adobe.com/useId). There is one dimension to this change that gives me pause: in a React v16 or v17 project using SSR, this approach requires the use of RA's [SSRProvider](https://react-aria.adobe.com/SSRProvider). Curious what your thoughts on the tradeoffs here are?

@mkernohanbc I've been giving this a poke and I think this trade-off is safe for us to make to get the benefits of React Aria's `useId()` here.

I have a repo here for collecting B.C. Design System stats using the `gh` CLI tool and `react-scanner` to analyze component usage: [ty2k/design-system-stats](https://github.com/ty2k/design-system-stats)

If you look at the [summary statistics doc](https://github.com/ty2k/design-system-stats/blob/main/summary.md#react-versions), you can see that this repository is the only one in `bcgov` that uses the React components library and mentions a version of React below v18. I'm going to update that stats repo further to generate a table of all React versions in use (including apps not using our components), but I'm getting rate-limited by GitHub.

My instinct here is that if we're not going to cause known adopters a headache by doing this, we should go ahead and do it.

Can you please update the README's "Supported React versions" section with a note for React 16 and 17 users to point them to the React Aria [SSRProvider doc page](https://react-aria.adobe.com/SSRProvider)?

ty2k (approved):
Great work @mkernohanbc, feel free to update the README in this PR or another one.

PR #762, opened 2026-06-26
Test new publish workflows

mkernohanbc:
This PR creates a temporary testbed for some proposed changes to the GHAs that publish new versions of `@bcgov/design-tokens` and `@bcgov/design-system-react-components`. For now, it does not touch these files directly. Instead, it creates two new workflows for POC:

- `test_publish_react_components`
- `test_publish_design_tokens`

## Proposed workflow changes
The goal of these proposed workflow changes is to significantly reduce redundancies in the build and publish process, by removing duplicated steps. This should make these workflows run significantly faster and consume fewer action minutes.

Currently, each workflow contains 4 separate jobs. This change moves duplicated steps between these 4 jobs into a new precursor `build` job, which precedes and is required by any of the publish jobs.

For example, the `design-system-react-components` build job handles the following steps (most of which are currently repeated independently in each publish job):

1. Check out code
2. Set up node
3. Install dependencies
4. Update version number (for `next` versions only)
5. Build components
6. Upload a build artifact via `actions/upload-artifact`

The publish jobs are thus simplified to:

1. Get the artifact created in `build` via `actions/download-artifact`
2. Set up node for their specific target environment
3. Publish

## Test workflows
The purpose of these test workflows is to enable us to test that this pattern works as expected. They remove most of the conditional logic for publishing, and instead use `workflow_dispatch` to enable us to mock a `next` or `latest` publish. The `--dry-run` flag is appended to all `npm publish` steps.

### DIFF
--- a/.github/workflows/publish_design_tokens.yaml
+++ b/.github/workflows/publish_design_tokens.yaml
@@ -10,34 +10,33 @@ on:
     types: [published]
 
 jobs:
-  publish-design-tokens-npm-next:
-    name: Publish @bcgov/design-tokens@next to npm
-    # Publish a `next` version to npm on PR merge to main branch.
-    if: ${{ github.event.pull_request.merged == true && github.repository == 'bcgov/design-system' }}
+  build:
+    if: ${{ (github.event.pull_request.merged == true && github.repository == 'bcgov/design-system') || (github.event_name == 'release' && contains(github.event.release.tag_name, 'bcgov/design-tokens')) }}
+    runs-on: ubuntu-latest
     permissions:
       contents: read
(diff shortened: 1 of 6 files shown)


ty2k (approved):
No red flags from me, seems like a big efficiency win. Up to you whether you want to keep the Copilot nit picks. ✅

mkernohanbc:
> Up to you whether you want to keep the Copilot nit picks. ✅

I'm going to disregard the `concurrency` stuff for now — it's unnecessary for the purposes of this testing. I'll circle back on it if/when we move forward with making these changes in the prod workflows.

PR #763, opened 2026-06-26
Test workflows: set correct download path for build artifacts

mkernohanbc:
Attempt to fix this job failure in the test workflows from #762: https://github.com/bcgov/design-system/actions/runs/28268823077/job/83761783124


ty2k (approved):
🚀

PR #772, opened 2026-07-02
TimeField: properly handle isRequired prop

mkernohanbc:
This PR fixes a mistake in how the `isRequired` prop is used in TimeField that surfaced during some testing work.

Previously, `isRequired` was destructured in order to render the visible "(required)" label — but this silently broke the actual required-input functionality, resulting in `data-required="true"` not being set on the TimeField.

This change removes this explicit destructuring, and instead properly passes it as a render prop (matching the pattern in other components) to handle the display of the label. With this change, `data-required="true"` is now also properly set on the TimeField.


ty2k (approved):
✅

PR #773, opened 2026-07-02
NumberField: remove hardcoded minValue

mkernohanbc:
This change corrects an error in how `minValue` is handled in NumberField, that surfaced during some testing work.

NumberField previously had its `minValue` hardcoded to `0` — which is the desired default behaviour, but would also potentially cause issues if someone wanted to set a different `minValue`.

This change destructures `minValue` in order to set the desired default, without hard-coding it.


ty2k (approved):
✅

PR #822, opened 2026-08-06
Fix transitive dependency vulnerability

mkernohanbc:
This PR patches a vulnerability (see GHSA-3jxr-9vmj-r5cp) in one of our transitive dependencies via `npm audit fix`.

(I mistakenly thought this was fixed by #805 and #806.)


ty2k (approved):
@mkernohanbc Looks good!

PR #857, opened 2026-08-28
Development principles for libraries

ty2k:
I've roughed out some dev principles for the design tokens and React components. Most of this is stuff we've only talked through verbally or in issues/PRs. I figure the READMEs are as good a place as any to start. Let me know what you think @mkernohanbc.


trev-dev (approved):

mkernohanbc (approved):
Love this, thanks for putting it together @ty2k! I think this is a great starting point, and we can refine it as we go.

